PR #17, opened 2026-03-04
MGMT-23016: management cluster upgrade implementation

maorfr:
part of https://issues.redhat.com/browse/MGMT-23016

related to https://github.com/gori-project/GoRI/issues/797 (https://github.com/gori-project/GoRI/issues/605)

this PR adds tasks to perform an upgrade of the management cluster openshift version.
the tasks are executed as part of the day2 operations playbook, which is the expected MSP's interface for content update related tasks.

the cluster upgrade process is based on https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html-single/updating_clusters/index#updating-cluster-cli.

the tasks are resilient to the process being interrupted and restarted.

note: etcd backup is not included in this PR (pending input and/or may be implemented in a follow up PR).

## Summary by CodeRabbit

* **New Features**
  * Added a management-cluster upgrade workflow for Day‑2 operations: automated OpenShift version upgrades with downgrade protection, validation of mirrored images and target availability, gating on operator health and upgradeability, conditional execution only when safe, and monitored completion with retries/polling for reliable, observable upgrades.

### DIFF
--- a/playbooks/tasks/mgmt_cluster_upgrade.yaml
+++ b/playbooks/tasks/mgmt_cluster_upgrade.yaml
@@ -28,7 +28,7 @@
 - name: Get available upgrades
   when: upgrade_required
   ansible.builtin.set_fact:
-    available_versions: "{{ cv_info.resources[0].status.availableUpdates | map(attribute='version') | list }}"
+    available_versions: "{{ (cv_info.resources[0].status.availableUpdates | default([])) | map(attribute='version') | list }}"
 
 - name: Determine if version is available (upgrade condition)
   when:
@@ -77,15 +77,20 @@
       }}
 
 - name: Determine if need to perform upgrade (aggregate upgrade conditions)
+  when: upgrade_required
   ansible.builtin.set_fact:
     perform_upgrade: >
       {{
-        upgrade_required
-        and upgrade_condition_version_mirrored
+       upgrade_condition_version_mirrored
         and upgrade_condition_version_available
         and upgrade_condition_co_healthy
         and upgrade_condition_co_upgradeable
       }}
+
+- name: Set perform_upgrade to false when upgrade not required
+  when: not upgrade_required
+  ansible.builtin.set_fact:
+    perform_upgrade: false
 
 - name: Trigger cluster upgrade
   when: perform_upgrade
--- a/playbooks/tasks/mgmt_cluster_upgrade.yaml
+++ b/playbooks/tasks/mgmt_cluster_upgrade.yaml
@@ -14,6 +14,14 @@
 - name: Determine if upgrade is required
   ansible.builtin.set_fact:
     upgrade_required: "{{ mgmt_openshift_version is ansible.builtin.version(cluster_version, '>') }}"
+    downgrade_requested: "{{ mgmt_openshift_version is ansible.builtin.version(cluster_version, '<') }}"
+
+- name: Fail when requested version is lower than current cluster version
+  when: downgrade_requested
+  ansible.builtin.fail:
+    msg: >-
+      Requested version {{ mgmt_openshift_version }} is lower than current cluster version {{ cluster_version }}.
+      Downgrades are not supported.
 
 - name: Get mirrored versions
   when: upgrade_required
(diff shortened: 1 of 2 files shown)


playbooks/tasks/mgmt_cluster_upgrade.yaml
```diff
@@ -0,0 +1,113 @@
...
+  kubernetes.core.k8s_info:
+    api_version: config.openshift.io/v1
+    kind: ClusterVersion
+    name: version
+  register: cv_info
+  until: >
+    cv_info.resources[0].status.history[0].state == 'Completed' and
+    cv_info.resources[0].status.history[0].version == mgmt_openshift_version
+  retries: 120
+  delay: 60
```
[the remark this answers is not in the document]
maorfr: we want to run this task without a condition. the reason is in case the execution was interrupted and the playbook is executed again - we want it to avoid performing the upgrade but still wait for the upgrade to complete.

playbooks/tasks/mgmt_cluster_upgrade.yaml
```diff
@@ -0,0 +1,118 @@
...
+       upgrade_condition_version_mirrored
+        and upgrade_condition_version_available
+        and upgrade_condition_co_healthy
+        and upgrade_condition_co_upgradeable
+      }}
+
+- name: Set perform_upgrade to false when upgrade not required
+  when: not upgrade_required
+  ansible.builtin.set_fact:
+    perform_upgrade: false
```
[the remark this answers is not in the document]
maorfr: in case the execution was interrupted and the playbook is executed again - we want it to avoid performing the upgrade but still wait for the upgrade to complete.

playbooks/tasks/mgmt_cluster_upgrade.yaml
```diff
@@ -0,0 +1,137 @@
...
+    api_version: config.openshift.io/v1
+    kind: ClusterVersion
+    name: version
+    definition:
+      spec:
+        desiredUpdate:
+          version: "{{ mgmt_openshift_version }}"
+          force: false
+
+- name: Wait for cluster upgrade to complete
```
eurijon, line 121: The task will be running unconditionally, don't we want to wait only after the upgrade is performed? (`when: perform_upgrade`). Otherwise the task will run (and pass) but might confuse the user.

playbooks/tasks/mgmt_cluster_upgrade.yaml
```diff
@@ -0,0 +1,137 @@
...
+- name: Get current desired cluster version
+  ansible.builtin.set_fact:
+    cluster_version: "{{ cv_info.resources[0].status.desired.version }}"
+
+- name: Determine if upgrade is required
+  ansible.builtin.set_fact:
+    upgrade_required: "{{ mgmt_openshift_version is ansible.builtin.version(cluster_version, '>') }}"
+    downgrade_requested: "{{ mgmt_openshift_version is ansible.builtin.version(cluster_version, '<') }}"
+
+- name: Fail when requested version is lower than current cluster version
```
eurijon, line 22: What if the version is the same, should we fail in that case too?

eurijon (commented):
Just a thought about variable location (and not directly related to this PR) - if the user is going to change `mgmt_openshift_version` from [defaults/platforms.yaml](https://github.com/rh-ecosystem-edge/enclave/blob/d629a0d900623a8bc2be20fea1dd770afa6abda5/defaults/platforms.yaml), should it be placed in a more user-facing file?

playbooks/tasks/mgmt_cluster_upgrade.yaml
maorfr: this is in case execution is interrupted. it allows to re-run the tasks and pick up where we left off, even without actually performing the patch again.

playbooks/tasks/mgmt_cluster_upgrade.yaml
maorfr: same as above, the same version is valid input to handle failure scenarios and re-running the tasks.

maorfr:
to answer the top level question - the user doesn't get to choose the management cluster version, it is decided (and thoroughly tested) by the partner.

eurijon (approved):

PR #25, opened 2026-03-04
fix: Enable Partner Overlay deployment

pgodowski:
The `bootstrap.sh` has disabled partner overlay install script call.

## Release Notes

* **Chores**
  * Partner Overlay deployment is now enabled and active during system deployment. The installation and startup routines will execute as part of the standard deployment process.


maorfr:
perhaps it would make sense to remove from this repository the partner-install directory, and execute it conditionally only if it exists. wdyt? this would make it suitable for testing and non-testing purposes.

pgodowski:
makes a lot of sense

maorfr:
let us know if you prefer us to implement! :pray:

pgodowski:
I will let you implement. If you confirm , feel free to close this PR

maorfr:
implementing in #29

PR #29, opened 2026-03-05
start partner install conditionally only if file exists

maorfr:
related to https://github.com/gori-project/GoRI/issues/816

replaces #25

in this PR we uncomment the partner install start script execution and make the execution conditional only if the file exists. this will facilitate testing and production needs.

## Summary by CodeRabbit

* **Chores**
  * Deployment wrapper now performs a runtime check before attempting partner installation and provides clear progress and fallback messages when the installer is absent.
  * Removed an obsolete placeholder installation script that was no longer used.


bootstrap.sh
```diff
@@ -151,10 +151,14 @@ printf '%b\n' "$(tail -3 ${workingDir}/ocp-cluster/.openshift_install.log \
   | sed -e 's/^"//' -e 's/"$//' -e 's/\\n/\
 /g' -e 's/\\"/"/g')"
 
-#echo -p "Deploying Partner OverLay .. " -n1 -s
-#    ./partner-install/start.sh ${workingDir}/ocp-cluster/auth/kubeconfig ${global_vars} ${certs_vars} 2>&1 >> ${log}
-#echo -e "\e[38;5;10m Done...\033[0m"; date
-#
+echo -p "Deploying Partner OverLay .. " -n1 -s
+    if [ -f ./partner-install/start.sh ]; then
+        ./partner-install/start.sh ${workingDir}/ocp-cluster/auth/kubeconfig ${global_vars} ${certs_vars} 2>&1 >> ${log}
```
kbsingh: is it worth checking the script is +x, or maybe force it with a /usr/bin/bash ?

kbsingh: minor other point would be if the global_vars and certs_var's are absolute references (ie, consuming when needed the working_dir prefix )

kbsingh (commented):
just two minor comments, looks good otherwise

bootstrap.sh
maorfr: executing the script with bash. the absolute reference to vars i'll keep for a future PR.

PR #32, opened 2026-03-05
Fail playbook when Ironic node reaches deploy failed state

eurijon:
Stop silently accepting 'deploy failed' as a successful deployment outcome.
Add `failed_when` so the task fails immediately when the node reaches that state, while still exiting the retry loop early.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved failure detection during hardware node provisioning: the wait step now aborts immediately if a node enters a "deploy failed" state, preventing further retries.
  * Preserves the existing until/retries/delay retry loop while providing faster failure feedback and clearer error behavior during provisioning.

### DIFF
--- a/playbooks/tasks/configure_hardware_ironic_node.yaml
+++ b/playbooks/tasks/configure_hardware_ironic_node.yaml
@@ -189,6 +189,7 @@
     status_code: [200]
   register: node_state
   until: node_state.json.provision_state in ["active", "deploy failed"]
+  failed_when: node_state.json.provision_state == "deploy failed"
   retries: 60
   delay: 10
   when: current_provision_state == "available"


eurijon:
@agonzalezrh this change as it's now it's failing the playbook when the node goes to `deploy failed`, meaning that it exists from the `agent_hosts` loop and doesn't execute [playbooks/tasks/wait_for_deployment.yaml](https://github.com/rh-ecosystem-edge/enclave/blob/1f4b1b32d39e3a0cb7255ea7274da4768ee6ce0a/playbooks/tasks/wait_for_deployment.yaml), so [playbooks/tasks/configure_hardware_ironic_detach_vmedia.yaml](https://github.com/rh-ecosystem-edge/enclave/blob/1f4b1b32d39e3a0cb7255ea7274da4768ee6ce0a/playbooks/tasks/configure_hardware_ironic_detach_vmedia.yaml) and [playbooks/tasks/configure_hardware_ironic_cleanup.yaml](https://github.com/rh-ecosystem-edge/enclave/blob/1f4b1b32d39e3a0cb7255ea7274da4768ee6ce0a/playbooks/tasks/configure_hardware_ironic_cleanup.yaml) are not executed. Is that ok? Should we do a proper cleanup if there is some issue with a node?

agonzalezrh (approved):

PR #33, opened 2026-03-05
update documentation

oourfali:
As we are providing discovery only through ACM,
we update the documentation to explain the enclave approach is a one-time installation convenience. Any changes should be done via ACM. And in general it is recommended to configure all nodes via ACM to begin with.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance recommending Red Hat ACM as the preferred approach for production bare metal host discovery and lifecycle management.
  * Clarified that Enclave configuration is a one-time setup option unsuitable for ongoing host management.
  * Restructured deployment documentation with improved operational pathways and clearer prerequisites for discovery workflows.


oourfali:
CC @agonzalezrh

agonzalezrh (approved):
lgtm, thanks

PR #58, opened 2026-03-09
Add retries to kubernetes.core.k8s calls

eurijon:
Add a retry mechanism to `kubernetes.core.k8s` `present`, `absent`, and `patched` calls to make them more robust against transient API or network issues.

Define default retries and delay values in `defaults/k8s.yaml`:
 - `k8s_retries: 6`
 - `k8s_delay: 5`

MGMT-23221

## Release Notes

* **New Features**
  * Added configurable Kubernetes retry settings with new `k8s_retries` (default: 6) and `k8s_delay` (default: 5) configuration parameters.
  
* **Improvements**
  * Enhanced reliability across multiple operators and playbooks by implementing retry logic and success verification for Kubernetes operations, ensuring tasks complete successfully before proceeding to the next step.

### DIFF
--- a/defaults/k8s.yaml
+++ b/defaults/k8s.yaml
@@ -1,0 +1,5 @@
+---
+# Retry settings for kubernetes.core.k8s state: present/absent/patched calls.
+# These are internal defaults and are not meant to be overridden by the user.
+k8s_retries: 6
+k8s_delay: 5
--- a/operators/advanced-cluster-management/acm_cis.yaml
+++ b/operators/advanced-cluster-management/acm_cis.yaml
@@ -25,6 +25,10 @@
   loop: "{{ r_cluster_image_set.resources }}"
   loop_control:
     loop_var: imageset
+  register: r_delete_clusterimageset
+  retries: "{{ k8s_retries }}"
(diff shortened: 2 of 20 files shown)


maorfr:
i think it's a bit of a wide net to throw. not every k8s operation should be retried blindly.

is there a particular place or places you are seeing this?

perhaps consider to add the defaults, but only add the actual retries to where we see failures?

eurijon:
> i think it's a bit of a wide net to throw. not every k8s operation should be retried blindly.
> 
> is there a particular place or places you are seeing this?
> 
> perhaps consider to add the defaults, but only add the actual retries to where we see failures?

It's an improvement to make the k8s calls more robust in general. I think @agonzalezrh has experienced issues when deploying in bare metal. I'm ok reducing them to those calls where we see issues, but I'm lacking that info.

agonzalezrh (approved):

agonzalezrh:
yes, reason ansible and in this case kubernetes.core.k8s are not good handling errors, if for any reason the call the api fails (i.e. pod is restarting), it is not retrying and all the workflow fails,

PR #70, opened 2026-03-10
test disconnected install with quay in non-latest version

maorfr:
testing lower operator version following https://github.com/rh-ecosystem-edge/enclave/pull/56

the assumption is that the issue is with mirroring.


maorfr:
testing disconnected deployment: https://github.com/rh-ecosystem-edge/enclave/actions/runs/22902967530

PR #77, opened 2026-03-11
adding ansible-automation-platform-24 and ansible-automation-platform-26 to registries.conf

agonzalezrh:
Adding all channels to the AAP, requires to synchronize 2.4 and 2.6


maorfr (approved):

maorfr:
should also update https://github.com/rh-ecosystem-edge/enclave/blob/main/operators/multicluster-engine/tasks.yaml?

PR #79, opened 2026-03-11
fix: Block 3scale images (tech preview) and update operators version

agonzalezrh:
## Summary by CodeRabbit

* **Chores**
  * Added three Authorino-related container images to the blocked images list.
  * Updated operator versions for Red Hat OADP and MetalLB to newer releases.

### DIFF
--- a/defaults/operators.yaml
+++ b/defaults/operators.yaml
@@ -62,7 +62,7 @@ operators:
     global: true
 
   - name: redhat-oadp-operator
-    version: 1.5.4
+    version: 1.5.5
     channel: stable
     namespace: openshift-oadp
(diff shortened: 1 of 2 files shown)


maorfr (dismissed):

defaults/operators.yaml
```diff
@@ -97,7 +97,7 @@ operators:
     source: cs-redhat-operator-index-v4-20
 
   - name: metallb-operator
-    version: 4.20.0-202602091941
+    version: metallb-operator.v4.20.0-202602261925
```
maorfr, line 100: I think this is wrong?

maorfr: you want to use csvName

PR #83, opened 2026-03-11
Add skip cleanup option to nightly workflows

eliorerz:
Add optional skip-cleanup checkbox when running nightly workflows manually. This allows keeping infrastructure running after deployment for debugging or manual testing.

Changes:
- Add skip-cleanup input parameter to both nightly workflows
- Cleanup step now checks if skip-cleanup is enabled
- Add warning notice when cleanup is skipped with cleanup instructions
- Scheduled runs always clean up regardless of setting

## Summary by CodeRabbit

* **Chores**
  * Added a skip-cleanup option to nightly E2E workflows. When run manually, users can skip automatic infrastructure cleanup; scheduled runs still clean up automatically.
  * When cleanup is skipped, the workflow now posts a clear notice with manual cleanup instructions and the cluster identifier.

### DIFF
--- a/.github/workflows/nightly-e2e-connected.yml
+++ b/.github/workflows/nightly-e2e-connected.yml
@@ -15,6 +15,11 @@ on:
         required: false
         type: boolean
         default: false
+      skip-cleanup:
+        description: 'Skip cleanup after deployment (leave infrastructure running)'
+        required: false
+        type: boolean
+        default: false
 
 # Prevent concurrent nightly runs
 concurrency:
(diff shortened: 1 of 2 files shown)


eliorerz:
The issue identified in the review has already been addressed. The commit was amended after the initial push to include the conditional cleanup logic in both workflows.

**Current state (commit f1b896c):**
- ✅ Both workflows declare the `skip-cleanup` input parameter
- ✅ Both workflows use the input in the cleanup step condition: `if: always() && (github.event_name == 'schedule' || inputs.skip-cleanup \!= true)`
- ✅ Both workflows include the "Cleanup skipped notice" step when cleanup is skipped
- ✅ Scheduled runs always clean up regardless of the input
- ✅ Manual runs respect the checkbox setting

The review comment was based on commit 0776cf7 (before amend) which had the bug, but the current PR state is correct.

PR #86, opened 2026-03-12
mirror single operator channel with min and max version

maorfr:
following up on #72

this PR reverts the above to mirror a single channel per operator, and adds a minVersion to start mirroring from. we also add maxVersion to mirror to.

this should limit the size of the content mirrored while explicitly defining the desired content to mirror without relying to oc-mirro defaults which are not functioning as expected.

## Summary by CodeRabbit

* **New Features**
  * Operators now record an explicit initial-install version to improve install/version resolution.
  * Image set templates now include per-operator default channel and channel ranges for more precise version selection.

* **Configuration Updates**
  * Added initial-version entries for many operators (including storage and model operators) and aligned channel/version metadata.
  * Network observability operator package name configured for consistency.

### DIFF
--- a/templates/imagesetconfiguration.yaml.j2
+++ b/templates/imagesetconfiguration.yaml.j2
@@ -14,12 +14,22 @@ mirror:
     - catalog: registry.redhat.io/redhat/certified-operator-index:v4.20
       packages:
         - name: gpu-operator-certified
+          defaultChannel: v25.10
+          channels:
+            - name: v25.10
+              minVersion: 25.10.0
+              maxVersion: 25.10.1
     - catalog: registry.redhat.io/redhat/redhat-operator-index:v4.20
       packages:
         - name: kubevirt-hyperconverged
         - name: kubernetes-nmstate-operator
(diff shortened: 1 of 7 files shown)


templates/imagesetconfiguration.yaml.j2
```diff
@@ -20,6 +20,11 @@ mirror:
         - name: kubernetes-nmstate-operator
 {% for operator in operators + model_operators %}
         - name: {{ operator.name }}
+          defaultChannel: {{ operator.channel }}
+          channels:
+            - name: {{ operator.channel }}
+              minVersion: {{ operator.min_version }}
```
agonzalezrh: @maorfr can you add a condiiton here, or we update the model operators too? before the split

agonzalezrh (changes requested):

templates/imagesetconfiguration.yaml.j2
maorfr: i split the sections

PR #89, opened 2026-03-12
Set defaults for RadosGW chunk size parameters

eurijon:
Set `minimum_chunk_size_mb` (100) and `maximum_chunk_size_mb` (500) as defaults in `quayBackendRGWDefaults`, with `server_side_assembly` always enabled.

Integer validation for both parameters is added to `validations.sh` when the values are overridden in `config/global.yaml`.

References:
- https://docs.redhat.com/en/documentation/red_hat_quay/3/html-single/configure_red_hat_quay/index#config-fields-storage-rados
- https://docs.redhat.com/en/documentation/red_hat_quay/3/html/manage_red_hat_quay/preparing-registry-large-artifacts

MGMT-23416

## Summary by CodeRabbit

* **New Features**
  * Added optional chunk-size settings for RadosGW storage: minimum_chunk_size_mb (default 100 MB) and maximum_chunk_size_mb (default 500 MB).
  * Enabled server-side assembly for RadosGW operations.
  * Added runtime validations to ensure chunk-size parameters are integers when provided.

* **Documentation**
  * Updated configuration reference and deployment guidance with examples, defaults, and notes for the new RadosGW parameters.

### DIFF
--- a/config/global.example.yaml
+++ b/config/global.example.yaml
@@ -45,6 +45,9 @@ quayBackendRGWConfiguration:
   secret_key: YOUR_S3_SECRET_KEY
   bucket_name: YOUR_S3_BUCKET_NAME
   hostname: YOUR_RADOSGW_ENDPOINT_HOSTNAME
+  # Uncomment only to override
+  # minimum_chunk_size_mb: YOUR_MIN_CHUNK_SIZE_MB  # Default: 100
+  # maximum_chunk_size_mb: YOUR_MAX_CHUNK_SIZE_MB  # Default: 500
 
 # Option 2: Local storage (NOT recommended for production)
 # quayBackend: LocalStorage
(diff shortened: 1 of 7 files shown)


maorfr:
related to https://github.com/gori-project/GoRI/issues/813

config/global.example.yaml
```diff
@@ -45,6 +45,9 @@ quayBackendRGWConfiguration:
   secret_key: YOUR_S3_SECRET_KEY
   bucket_name: YOUR_S3_BUCKET_NAME
   hostname: YOUR_RADOSGW_ENDPOINT_HOSTNAME
+  # Uncomment only to override
+  # minimum_chunk_size_mb: YOUR_MIN_CHUNK_SIZE_MB  # Default: 100
+  # maximum_chunk_size_mb: YOUR_MAX_CHUNK_SIZE_MB  # Default: 500
```
maorfr, line 50: i'm not sure we need these in config. if someone REALLY wants to override let them do it in defaults?

maorfr (approved):

PR #100, opened 2026-03-15
add option to enable/disable openshift-ai and nvidia-gpu

maorfr:
part of:
- https://issues.redhat.com/browse/MGMT-23443
- https://issues.redhat.com/browse/MGMT-23442

in this PR we are adding 2 variables to the global configuration file:
- `enable_openShift_ai` - boolean, false by default
- `enable_nvidia_gpu` - boolean, false by default

the bootstrap process will include or not the required controls that implement these functionalities:
- mirror nvidia content
- apply acm policies

these functionalities may be enabled in day2 by updating the variables in the global configuration file and running `sync.sh`, which is the already established day2 entrypoint.

## Summary by CodeRabbit

* **New Features**
  * Added enable_openshift_ai and enable_nvidia_gpu flags to toggle OpenShift AI and NVIDIA GPU components (both default to disabled).
  * AI/GPU-related components and image entries are included only when their flags are enabled.

* **Documentation**
  * Deployment guide updated with the new configuration options and certificate formatting guidance.

* **Changed**
  * Changelog updated to reference the new feature flags.

### DIFF
--- a/templates/imagesetconfiguration.yaml.j2
+++ b/templates/imagesetconfiguration.yaml.j2
@@ -11,6 +11,7 @@ mirror:
 {% endfor %}
     graph: true
   operators:
+{% if enable_nvidia_gpu | default(false) %}
     - catalog: registry.redhat.io/redhat/certified-operator-index:v4.20
       packages:
         - name: gpu-operator-certified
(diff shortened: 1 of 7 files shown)


templates/imagesetconfiguration.yaml.j2
```diff
@@ -56,6 +58,7 @@ mirror:
     - name: {{ osac_operator.image }}:{{ osac_operator.tag }}
     - name: {{ aap_bootstrap_job.image }}:{{ aap_bootstrap_job.tag }}
 
+{% if enable_openshift_ai | default(false) %}
```
danielerez, line 61: just in case we'd want to block images unrelated to AI, maybe worth moving the condition to line 63?
or, just remove the condition, as it shouldn't really affect the mirroring.

maorfr: once we have more images we will move the if one line down

danielerez: > once we have more images we will move the if one line down

ok. hope it won't be a bit confusing though, as the list is quite large:/

PR #102, opened 2026-03-15
conditional call to partner validations

maorfr:
similar to bootstrap.sh calling partner install, this PR adds a call from validations.sh to partner validations.

this adds a partner overlay to 2 of our 3 entrypoints:
1. bootstrap.sh
2. validations.sh

the 3rd entrypoint is sync.sh <-- food for thought.

## Summary by CodeRabbit

* **New Features**
  * Optional partner-specific validations are now supported: when present they run automatically and their output is appended to system logs; when absent the system reports that partner validations were skipped. Existing validation behavior is preserved when no partner validations are provided.

### DIFF
--- a/validations.sh
+++ b/validations.sh
@@ -384,3 +384,9 @@ else
 fi
 
 # END OF VALIDATIONS
+
+if [ -f ./partner-install/validations.sh ]; then
+    bash ./partner-install/validations.sh
+else
+    echo "Partner validations not found, skipping"
+fi


maorfr:
related to https://github.com/gori-project/GoRI/issues/827

validations.sh
```diff
@@ -384,3 +384,9 @@ else
 fi
 
 # END OF VALIDATIONS
+
+if [ -f ./partner-install/validations.sh ]; then
+    bash ./partner-install/validations.sh 2>&1 >> ${log}
```
danielerez: should we handle an undefined 'log' var?
also, do we want to print errors to screen here, or is it just a typo?
i.e. to redirect stderr to log file, '2>&1' should be at EOL

kbsingh:
@maorfr lets work through how best to implement these hooks, and what the integrations might look like.

PR #106, opened 2026-03-17
disable quay feature quota management

maorfr:
test mirror speed without quota management.

## Summary by CodeRabbit

* **Chores**
  * Updated default configuration for the quota management feature. Quota management is now disabled by default, enabling more flexible deployments and simplified setup for users who don't require this capability. Administrators and operators can enable quota management through their configuration settings as needed for their specific deployment requirements and use cases.


maorfr:
https://github.com/gori-project/GoRI/issues/828

PR #113, opened 2026-03-17
Fix duplicate platform.channels entries in imagesetconfiguration template

rporres:
The previous template emitted one channel entry per item in openshift_versions. When multiple versions share the same minor (e.g. 4.20.8 and 4.20.10), this produced duplicate stable-4.20 entries that oc-mirror rejects.

Replace the simple per-version loop with a two-pass approach:
- First pass: group versions by channel name using a dict, tracking the minimum and maximum patch level per channel. Patch levels are compared as integers, so 10 > 8 rather than "10" < "8" lexicographically.
- Second pass: emit one channel entry per unique channel with the correct minVersion and maxVersion.

Insertion order of channels in the output matches the order in which each minor version is first encountered in openshift_versions.

## Summary by CodeRabbit

* **Refactor**
  * Channel configuration system now dynamically aggregates based on version availability instead of using hard-coded entries, improving flexibility and accuracy in channel management.

### DIFF
--- a/templates/imagesetconfiguration.yaml.j2
+++ b/templates/imagesetconfiguration.yaml.j2
@@ -4,10 +4,25 @@ apiVersion: mirror.openshift.io/v2alpha1
 mirror:
   platform:
     channels:
+{% set channels = {} %}
 {% for v in openshift_versions %}
-      - name: stable-{{ v.version.split('.')[:2] | join('.') }}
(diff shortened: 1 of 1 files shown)


rporres:
At some point I'd really like to add tests for complex templates such as this one, but with the current planned changes I think it's enough for now :)

maorfr (approved):

PR #117, opened 2026-03-18
Add new signatures after the internal quay mirroring

rporres:
Without this, trying to upgrade a cluster after a sync from LZ to quay registry fails as the update cannot be verified.

This was found in the context of https://redhat.atlassian.net/browse/MGMT-23449

## Summary by CodeRabbit

* **Chores**
  * Added a deployment step to the Quay operator’s disconnected workflow to apply the release signature ConfigMap after registry configuration cleanup.
  * Added an environment override so Quay disconnected tasks run with the cluster kubeconfig context, aligning their execution environment with other Day‑2 disconnected/cluster configuration steps.

### DIFF
--- a/operators/quay-operator/quay_disconnected.yaml
+++ b/operators/quay-operator/quay_disconnected.yaml
@@ -49,3 +49,8 @@
   ansible.builtin.file:
     path: "{{ lookup('env','HOME') }}/.config/containers/registries.conf"
     state: absent
+
+- name: Apply release signature ConfigMap
+  ansible.builtin.command:
+    cmd: |
+      {{ workingDir }}/bin/oc apply -f {{ workingDir }}/config/oc-mirror-workspace-quay/working-dir/cluster-resources/signature-configmap.yaml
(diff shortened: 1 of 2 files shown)


maorfr (dismissed):

maorfr:
actually maybe this step belongs in mirror-registry?

maorfr:
https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/configure_ocp_abi.yaml#L57

rporres:
> actually maybe this step belongs in mirror-registry?
> https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/configure_ocp_abi.yaml#L57

It would be nice, but it looks like those manifests that are copied there are not applied as part of the mirror playbook (most likely because there's no mgmt cluster yet), so we would need to further hack things for this manifest.

~On top of that, I've compared signatures from registry and quay from my last sync and they're not exactly equal, I still need to understand why~ 🤷 

Both signature configmaps contain the same data.

maorfr (approved):

PR #119, opened 2026-03-18
MGMT-23405: add in-cluster Quay fallback mirrors and CA trust

danielerez:
Handle failover when the landing-zone (LZ) mirror registry is down or unreachable. In disconnected setups, nodes normally pull from the LZ Quay (mirror.<domain>:8443).
If that registry is stopped or unavailable, image pulls fail. This change adds the in-cluster Quay as a fallback so that IDMS/ITMS direct registry.redhat.io (and other sources) to the internal Quay when the LZ mirror cannot be reached, keeping pulls working.

quay_disconnected.yaml:
- After oc-mirror and registries.conf cleanup, include quay_disconnected_mirrors.yaml to apply in-cluster Quay mirrors and configure node trust for the Quay registry.

quay_disconnected_mirrors.yaml (new file):
- Copy oc-mirror-generated IDMS/ITMS to '-internal' variants and apply to the cluster so registry.redhat.io (and others) can fall back to the in-cluster Quay when the landing-zone mirror is unavailable. Coexists with landing-zone IDMS/ITMS.
- Add the ingress CA (router/Quay route cert) to image.config additionalTrustedCA so nodes trust the in-cluster Quay registry for image pulls and avoid x509 certificate errors. Creates or patches the ConfigMap and references it from image.config.openshift.io/cluster.

## Summary by CodeRabbit

* **New Features**
  * Automatically applies oc-mirror-generated image mirror manifests for disconnected Quay setups.
  * Creates internal mirror variants to avoid naming conflicts with landing-zone mirrors.
  * Falls back to internal Quay mirrors when landing-zone mirror configuration is unavailable.
  * Adds cluster trust configuration for the Quay registry TLS certificate so nodes can securely pull images; certificate fetch failures are non-fatal.
  * Ensures mirror and trust steps run as part of the Quay deployment flow.

### DIFF
--- a/operators/quay-operator/quay_disconnected.yaml
+++ b/operators/quay-operator/quay_disconnected.yaml
@@ -54,3 +54,8 @@
   ansible.builtin.command:
     cmd: |
       {{ workingDir }}/bin/oc apply -f {{ workingDir }}/config/oc-mirror-workspace-quay/working-dir/cluster-resources/signature-configmap.yaml
+
+# Apply oc-mirror-generated IDMS/ITMS to the cluster (fallback to internal Quay when LZ fails)
+# and add the Quay registry CA to image.config so nodes trust the registry for image pulls.
+- name: Apply Quay mirrors and trust registry CA for image pulls
+  ansible.builtin.include_tasks: quay_disconnected_mirrors.yaml
--- a/operators/quay-operator/quay_disconnected_mirrors.yaml
+++ b/operators/quay-operator/quay_disconnected_mirrors.yaml
@@ -1,0 +1,144 @@
+---
+# Quay disconnected: apply IDMS/ITMS from oc-mirror and trust Quay registry CA for image pulls.
+# Included from quay_disconnected.yaml.
+# Copy to idms-oc-mirror-internal.yaml / itms-oc-mirror-internal.yaml, rename the resource inside,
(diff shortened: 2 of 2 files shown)


rporres:
Are we interested in the order of resolution of both registries? I understand that we prefer to pull from in-cluster registry as there's a good chance that LZ is not available once the mgmt cluster is installed. I don't know how the order of resolution is going to work with the manifests from oc-mirror applied as-is...

danielerez:
> Are we interested in the order of resolution of both registries? I understand that we prefer to pull from in-cluster registry as there's a good chance that LZ is not available once the mgmt cluster is installed. I don't know how the order of resolution is going to work with the manifests from oc-mirror applied as-is...

Yeah, good point, I've considered exactly that. But I actually understood that we want to keep the in-cluster registry mirrors only as a fallback. I.e. We don't expect to completely lose availability of the LZ machine. So this solution is mainly addressing temporary issues. The in-cluster quay should still be the main registry for the spoke clusters, while the management cluster can keep using the LZ registry. We could of of course reconsider it later on, I think changing the order would just require update the names (alphabetical) of the idms/itms resources.

rporres (dismissed):
Fair enough!

danielerez:
Thanks @rporres :) Can you please approve again, just had to rebase...

PR #124, opened 2026-03-19
operator update change from single csvName to multiple csvNames and add csvNames for odf-operator

maorfr:
odf-operator installs dependencies using an operator called odf-dependencies.

when using a Subscription with installPlanApproval set to Manual, the dependent Subscriptions are also set to Manual. to approve them we explicit install them in advance (similar to the installation of multicluster-engine prior to advanced-cluster-management).

since this Subscription contains multiple CSVs we are modifying the operator update logic to support multiple csvNames. differently from multicluster-engine, in this case the installPlan is created in the same namespace, so it will suffice to add the list of csvNames to odf-operator itself.

## Summary by CodeRabbit

* **New Features**
  * ODF storage operator now bundles related storage components (Rook Ceph, OCS, MCG and companions) as part of its operator package.

* **Updates**
  * Operator configurations now accept multiple CSV identifiers, improving operator selection and subscription resolution during installs and updates.

### DIFF
--- a/defaults/operators.yaml
+++ b/defaults/operators.yaml
@@ -28,7 +28,8 @@ operators:
     init_version: 5.0.3
     namespace: openshift-update-service
     source: cs-redhat-operator-index-v4-20
-    csvName: update-service-operator
+    csvNames:
+      - update-service-operator
 
   - name: openshift-gitops-operator
     version: 1.19.2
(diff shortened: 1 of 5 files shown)


maorfr:
related to https://github.com/gori-project/GoRI/issues/833

javipolo (approved):

PR #127, opened 2026-03-19
MGMT-23598: Add plugin deployment infrastructure

eliorerz:
## Summary

- Plugin runner (`deploy-plugin.yaml`) with full lifecycle: pre-validate → mirror → operators → deploy → post-validate
- Reusable tasks file (`tasks/deploy_plugin.yaml`) for inclusion from other playbooks
- Shell wrapper (`deploy_plugin.sh`) for SSH-based deployment to Landing Zone
- Plugin validation script and CI job
- Core pipeline hooks: foundation plugin mirroring (Phase 2) and pre-install hardware validation (Phase 3)
- MCE ConfigMap patching for plugin registry entries in disconnected mode
- Example plugin as reference implementation
- Generic Makefile targets: `deploy-plugin`, `mirror-plugin`, `validate-plugins`
- `configure_ocp_abi.yaml`: replace hardcoded mirror manifest file list with fileglob pattern
- Fix: handle pre-existing Subscriptions with wrong CatalogSource in disconnected mode

## Plugin directory structure

```
plugins/<name>/
├── plugin.yaml                 # Required: name, type, order, mirror, operators
├── config/defaults.yaml        # Optional: plugin-specific variables
├── operators/operators.yaml    # Required if operators: true
├── mirror/
│   ├── imageset.yaml.j2        # Required if mirror: plugin or core
│   └── registries.yaml         # Optional: MCE ConfigMap entries
├── pre-install-validate.yaml   # Optional: hardware validation before install
├── pre-validate.yaml           # Optional: cluster readiness check
├── deploy.yaml                 # Optional: deployment tasks
└── post-validate.yaml          # Optional: post-deployment verification
```

## Test plan

- [x] `make validate` passes all checks
- [x] `make validate-plugins` validates example plugin
- [ ] No regressions on existing connected/disconnected CI

## Summary by CodeRabbit

* **New Features**
  * Full plugin framework: pre-validate, pre-install validation, deploy, post-validate; supports mirror modes, operator-managed plugins, foundation-plugin auto-deploy, and an example plugin.
  * Remote plugin deployment tooling and an ISO installation-gate injection to control install progress.
  * Enhanced orchestration: per-node boot handling and improved deployment/wait steps.

* **Tests**
  * CI plugin validation job: runs plugin checks, summarizes results, and uploads plugin logs on failure.

### DIFF
--- a/hack/agent-gate-inject.py
+++ b/hack/agent-gate-inject.py
@@ -1,0 +1,418 @@
+#!/usr/bin/env python3
+"""
+Inject an installation gate service into an agent-based installer ISO.
+
+This script modifies the ignition config embedded inside the ISO's
+/images/ignition.img so that start-cluster-installation.service is
+blocked until a signal file is created on the node.
+
+Usage:
+    python3 agent-gate-inject.py <agent.iso>
+
+To release the gate on the booted node:
+    touch /var/tmp/proceed-with-installation
+
+Requires: Python 3.6+  (no external dependencies)
+"""
+
+import argparse
+import gzip
+import io
+import json
+import struct
+import sys
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+ISO_SECTOR_SIZE = 2048
+# Primary Volume Descriptor starts at sector 16 (byte 32768)
+PVD_OFFSET = 16 * ISO_SECTOR_SIZE
+
+GATE_SCRIPT = """\
+#!/bin/bash
+echo "=========================================="
+echo "  Installation gate is ACTIVE"
+echo "  Waiting for release signal..."
+echo "  To proceed, run:"
+echo "    touch /var/tmp/proceed-with-installation"
+echo "=========================================="
+while [ ! -f /var/tmp/proceed-with-installation ]; do
+  sleep 5
+done
+echo "Gate released. Proceeding with installation..."
+"""
+
+GATE_SERVICE_UNIT = """\
+[Unit]
+Description=Installation gate - blocks start-cluster-installation
+Before=start-cluster-installation.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/installation-gate.sh
+RemainAfterExit=yes
+StandardOutput=journal+console
+
+[Install]
+RequiredBy=start-cluster-installation.service
+"""
+
+GATE_DROPIN = """\
+[Unit]
+Requires=installation-gate.service
+After=installation-gate.service
+"""
+
+
+# ---------------------------------------------------------------------------
+# ISO 9660 helpers  (read-only, just enough to locate files)
+# ---------------------------------------------------------------------------
+
+def read_iso9660_string(data, offset, length):
+    return data[offset:offset + length].decode("ascii", errors="replace").strip()
+
+
+def find_file_in_directory(iso_data, dir_offset, dir_length, target_name):
+    """Walk an ISO 9660 directory and return (offset, size) of target_name."""
+    pos = 0
+    while pos < dir_length:
+        record_len = iso_data[dir_offset + pos]
+        if record_len == 0:
+            # skip to next sector boundary
+            next_sector = ((pos // ISO_SECTOR_SIZE) + 1) * ISO_SECTOR_SIZE
+            if next_sector >= dir_length:
+                break
+            pos = next_sector
+            continue
+
+        extent_loc = struct.unpack_from("<I", iso_data, dir_offset + pos + 2)[0]
+        data_len = struct.unpack_from("<I", iso_data, dir_offset + pos + 10)[0]
+        name_len = iso_data[dir_offset + pos + 32]
+        name_raw = iso_data[dir_offset + pos + 33: dir_offset + pos + 33 + name_len]
+
+        # Strip version number (;1)
+        name = name_raw.decode("ascii", errors="replace").split(";")[0].strip(".")
+
+        if name.upper() == target_name.upper():
+            return extent_loc * ISO_SECTOR_SIZE, data_len
+
+        pos += record_len
+
+    return None, None
+
+
+def locate_iso_file(iso_data, path):
+    """Locate a file in the ISO by path (e.g. '/images/ignition.img').
+    Returns (byte_offset, byte_length)."""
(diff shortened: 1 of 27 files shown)


playbooks/tasks/pre_install_validate_plugins.yaml
```diff
@@ -0,0 +1,87 @@
+---
+- name: Get Assisted Service auth token from rendezvous node
+  ansible.builtin.shell: |
```
agonzalezrh, line 3: I still need to check all the PR, but this should be done using ansible.builtin.add_host

playbooks/tasks/configure_operator.yaml
```diff
@@ -225,7 +225,7 @@
         name: "{{ operator.name }}"
         source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
         sourceNamespace: openshift-marketplace
-        startingCSV: "{{ operator.csvName | default(operator.name) }}.v{{ operator.version | replace('+', '-') }}"
+        startingCSV: "{{ operator.csvNames | default([]) | first | default(operator.name) }}.v{{ operator.version | replace('+', '-') }}"
```
danielerez, line 228: seems it's already in main. just rebase?

playbooks/tasks/deploy_foundation_plugins.yaml
```diff
@@ -0,0 +1,45 @@
+---
+# Auto-discover and deploy all foundation plugins (type: foundation)
+# sorted by order field.
+#
+# Called from 05-operators.yaml at the end of Phase 5.
```
danielerez, line 5: maybe worth splitting it to a separate phase?

hack/agent-gate-inject.py
```diff
@@ -0,0 +1,418 @@
+#!/usr/bin/env python3
+"""
+Inject an installation gate service into an agent-based installer ISO.
```
danielerez, line 3: Why do we actually must use an installation gate?
I.e. Which validations are mandatory to invoke before installation? Can we do these validations simply before booting all machines? I mean, the installation won't start any way before all nodes are ready.

But if it's indeed mandatory, this logic seems a bit like an overkill:)
There's a much simple solution to hold installation in dev-scripts codebase: https://github.com/openshift-metal3/dev-scripts/blob/f047027b0dcff27e654eca45032031d1b8e8c793/agent/06_agent_create_cluster.sh#L315

It basically just disable the services that start installation. 
So for triggering installation manually later: 'sudo systemctl start start-cluster-installation.service'

playbooks/tasks/mirror_single_core_plugin.yaml
```diff
@@ -0,0 +1,56 @@
+---
```
danielerez, line 1: Do we need a separate file for core plugin? I.e. maybe we can reuse the generic deploy_plugin playbook with flags?

playbooks/tasks/deploy_plugin.yaml
```diff
@@ -0,0 +1,232 @@
...
+  ansible.builtin.include_vars:
+    file: "{{ plugin_dir }}/mirror/registries.yaml"
+    name: plugin_registries_file
+  when:
+    - _mirror_mode in ['core', 'plugin']
+    - disconnected | default(false) | bool
+    - r_plugin_registries_yaml.stat.exists
+  tags: mirror
+
+- name: Patch MCE custom-registries with plugin entries
```
danielerez, line 173: if it's specific for MCE, maybe we could extract this to plugin?

playbooks/tasks/deploy_plugin.yaml
```diff
@@ -0,0 +1,232 @@
...
+- name: Template plugin imageset configuration
+  ansible.builtin.template:
+    src: "{{ plugin_dir }}/mirror/imageset.yaml.j2"
+    dest: "{{ workingDir }}/plugin-{{ plugin_name }}-imageset.yaml"
+  when:
+    - _mirror_mode == 'plugin'
+    - disconnected | default(false) | bool
+    - lookup('ansible.builtin.fileglob', plugin_dir ~ '/mirror/imageset.yaml.j2') | length > 0
+  tags: mirror
+
```
eurijon, line 96: To avoid some misconfiguration at runtime maybe would be good to fail when `_mirror_mode in ['core', 'plugin'] and not disconnected`?

playbooks/deploy-plugin.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Deploy a plugin to the OpenShift cluster
+#
+# This playbook loads the plugin configuration, applies defaults,
+# and runs the plugin lifecycle: pre-validate -> mirror -> operators -> deploy -> post-validate.
+#
+# Usage:
+#   ansible-playbook playbooks/deploy-plugin.yaml -e workingDir=/home/cloud-user -e plugin_name=example
+#
+# Required variables:
+#   workingDir: Working directory on the Landing Zone
```
eurijon, line 11: Shouldn't be needed as far as `common/load-vars.yaml` is included (line 28), right? `workingDir` should be defined in `config/global.yaml`. There might be some case I'm missing...

maorfr:
this is a great start!

i'd like to recommend:
1. i like the idea of an interface for a plugin. i suggest to use a single file per plugin with well defined fields that encapsulate the functionality instead of the ansible implementation details. each plugin file should be schema validated (need to define schema properly).
2. i think we should clean some tech debt (like #137 for example) before introducing more fields into the interface.
3. i think the plugin system should have an entrypoint of its own, but at the same time it should trigger existing pieces of the process instead of introducing new steps.

this is just some initial feedback on the design. hope this helps!

oourfali:
1. I guess it's a trade-off with the level of freedom, vs being more strict. Do we think we need to be more strict in order to ease upgrade? or in general? 
2. @eliorerz how will https://github.com/rh-ecosystem-edge/enclave/pull/137 impact the plugin infra?

playbooks/tasks/pre_install_validate_plugins.yaml
eliorerz: The `delegate_to` approach fails because the rendezvous node is a minimal RHCOS env so the Ansible's SSH connection can't find a Python interpreter or establish a full connection the way it normally would. 
I think we can use `ansible.builtin.raw` but it would be less pretty

PR #130, opened 2026-03-20
[WIP] MGMT-23621: Extract LVMS and ODF storage operators into independent foundation plugins 

eliorerz:
Extract LVMS and ODF into separate storage plugins
Move LVMS and ODF storage operators from the core pipeline into
independent foundation plugins (plugins/lvms/, plugins/odf/).

Each plugin carries its own operator definition, deploy tasks,
mirror imageset, and registry entries. A new storage_backend: true
field in plugin.yaml marks mutually exclusive storage plugins.
The storage_plugin config variable (renamed from blockStorageBackend)
selects which one is deployed and mirrored.

Reorder 05-operators.yaml so foundation plugins deploy BEFORE core
operators, ensuring the storage backend is available when Quay and
other operators that depend on it are configured.

Add storage_backend filtering to deploy_foundation_plugins,
mirror_core_plugins, and pre_install_validate_plugins so only
the plugin matching storage_plugin is processed; all other storage
plugins are skipped.

Remove defaults/storage_operators.yaml, defaults/lvms_operator.yaml,
operators/lvms-operator/tasks.yaml, and schemas/lvms_operator.yaml.
Remove lvms4/odf4/rhceph entries from registries.conf.j2 and the
ODF conditional block from imagesetconfiguration.yaml.j2.

## Summary by CodeRabbit

* **New Features**
  * Plugin system: selectable storage/foundation plugins with discovery, ordered deployment, mirroring, pre-install/post-validate checks, an installation gate for agent installs, and remote plugin deploy command/playbook. Includes an example plugin.

* **CI / Tests**
  * Plugin Validation CI job and local validation command added; logs summarized in job output and failure artifacts uploaded with short retention.

* **Documentation**
  * Configuration and deployment docs updated to use a selectable storage_plugin and document plugin workflows.


playbooks/tasks/configure_operator.yaml
```diff
@@ -10,17 +10,26 @@
...
+    - r_og_check.resources | default([]) | length == 0
     - not (operator.global | default(false))
   kubernetes.core.k8s:
     state: present
     definition:
       apiVersion: operators.coreos.com/v1
       kind: OperatorGroup
       metadata:
-        name: "{{ operator.name }}-og"
+        name: "{{ operator.namespace }}-og"
```
maorfr, line 32: unrelated change

playbooks/tasks/configure_operator.yaml
```diff
@@ -33,14 +42,15 @@
...
+    - r_og_check.resources | default([]) | length == 0
     - operator.global | default(false)
   kubernetes.core.k8s:
     state: present
     definition:
       apiVersion: operators.coreos.com/v1
       kind: OperatorGroup
       metadata:
-        name: "{{ operator.name }}-og"
+        name: "{{ operator.namespace }}-og"
```
maorfr, line 53: unrelated change

playbooks/tasks/configure_operator.yaml
```diff
@@ -10,17 +10,26 @@
...
   until: r_namespace_exists is success
 
+- name: "Check if OperatorGroup already exists in namespace"
+  when: operator.namespace != "openshift-operators"
+  kubernetes.core.k8s_info:
+    api_version: operators.coreos.com/v1
+    kind: OperatorGroup
+    namespace: "{{ operator.namespace }}"
+  register: r_og_check
+
```
maorfr, line 20: not needed?

playbooks/tasks/configure_operator.yaml
```diff
@@ -56,10 +66,138 @@
...
+      register: r_csv
+      retries: "{{ k8s_retries }}"
+      delay: "{{ k8s_delay }}"
+      until:
+        - r_csv.resources[0].status.phase is defined
+        - r_csv.resources[0].status.phase == "Succeeded"
+      when:
+        - r_sub_source_cs.resources | length == 0
+        - r_correct_cs.resources | default([]) | length > 0
+
```
maorfr, line 196: this should be removed, all these aspects are handled elsewhere

playbooks/tasks/configure_operator.yaml
```diff
@@ -95,9 +234,9 @@
         installPlanApproval: Manual
         channel: "{{ operator.channel }}"
         name: "{{ operator.name }}"
-        source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
+        source: "{{ operator.source }}"
         sourceNamespace: openshift-marketplace
-        startingCSV: "{{ operator.csvNames | default([]) | first | default(operator.name) }}.v{{ operator.version | replace('+', '-') }}"
+        startingCSV: "{{ (operator.csvNames | default([]) | first | default(operator.name)) ~ '.v' ~ (operator.version | replace('+', '-')) }}"
```
maorfr, line 239: unrelated change

playbooks/tasks/configure_operator.yaml
```diff
@@ -95,9 +234,9 @@
         installPlanApproval: Manual
         channel: "{{ operator.channel }}"
         name: "{{ operator.name }}"
-        source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
+        source: "{{ operator.source }}"
```
maorfr, line 237: unrelated change

playbooks/tasks/configure_operator.yaml
```diff
@@ -83,6 +221,7 @@
   when:
     - r_sub_check.resources | length == 0
     - operator.namespace != "openshift-operators"
+    - disconnected | default(true) | bool
```
maorfr, line 224: unrelated change

playbooks/tasks/configure_operators.yaml
```diff
@@ -2,14 +2,9 @@
 - name: Configure operators
   block:
 
-    - name: Disable default operator catalog sources
-      when: disconnected | default(true) | bool
-      ansible.builtin.command: |
-        {{ workingDir }}/bin/oc patch OperatorHub cluster --type=json -p='[{"op":"replace","path":"/spec/disableAllDefaultSources","value":true}]'
```
maorfr, line 8: restore this please

playbooks/tasks/patch_mce_registries.yaml
```diff
@@ -0,0 +1,103 @@
+---
```
maorfr, line 1: this is no longer needed given we template registries.conf.j2

playbooks/tasks/collect_core_plugin_operators.yaml
```diff
@@ -0,0 +1,58 @@
+---
```
maorfr, line 1: no longer needed?

playbooks/tasks/collect_plugin_registries.yaml
```diff
@@ -0,0 +1,37 @@
+---
```
maorfr, line 1: no longer needed?

helpers/agent-gate/agent-gate-inject.py
```diff
@@ -0,0 +1,417 @@
+#!/usr/bin/env python3
```
maorfr, line 1: can installation gates be added in another PR after this one?

eliorerz: Basically yes, but it means that we will have race between the installation and the pre-install validations

playbooks/tasks/configure_operators.yaml
eliorerz: It wasn't removed, it was moved earlier in the pipeline see https://github.com/rh-ecosystem-edge/enclave/blob/31371d4210e9d16aa71fdd706043b5801ffe487d/playbooks/05-operators.yaml#L33

playbooks/tasks/configure_operator.yaml
eliorerz: Why not? without this, deploying a second operator into the same namespace fails because `k8s: state: present` for `OperatorGroup` with a different name fails, and two `OperatorGroups` in one namespace is invalid

playbooks/tasks/configure_operator.yaml
eliorerz: This is not directly related but without this, the OG check wouldn't help because each operator would try to create a differently-named OG.

playbooks/tasks/configure_operator.yaml
eliorerz: Same here https://github.com/rh-ecosystem-edge/enclave/pull/130/files#diff-39a7d34b5d3bd838b97881fcfd8416a72a092e9091f913b82eaf06bd71a35f86

playbooks/tasks/configure_operator.yaml
eliorerz: Foundation plugins with `source: set` need Manual approval in disconnected mode but should use Automatic in connected mode. Without this, connected deployments fail because there's no `CatalogSource` matching operator.source.

playbooks/tasks/configure_operator.yaml
eliorerz: The Manual Subscription path uses `operator.source` directly instead of the conditional `operator.source` if disconnected else 'redhat-operators'. This is correct because the Manual path now only runs in disconnected mode (related to https://github.com/rh-ecosystem-edge/enclave/pull/130/files#diff-39a7d34b5d3bd838b97881fcfd8416a72a092e9091f913b82eaf06bd71a35f86)

playbooks/tasks/configure_operator.yaml
eliorerz: ODF's package name is odf-operator but its CSV is also odf-operator, however other operators may diverge. The csvNames field was introduced as a part of the plugin system.

playbooks/tasks/configure_operator.yaml
eliorerz: This is a real scenario I had during testing. The `ingress-operator` auto-creates subscriptions with `source: redhat-operators` which doesn't exist in disconnected mode. Without this, those operators never install.
I don't think this is duplication (of the normal install flow) - it's handling a specific edge case

playbooks/tasks/patch_mce_registries.yaml
eliorerz: `registries.conf.j2` handles the hub cluster's container runtime config. MCE's custom-registries `ConfigMap` is a separate thing, it's used for spoke  cluster deployments. They serve different purposes.
If you still think that MCE registries are truly not needed in the current deployment model, it can be removed.

playbooks/tasks/collect_core_plugin_operators.yaml
eliorerz: This is the mechanism that gets plugin operators into the core `oc-mirror` run. Without it, plugin operators aren't mirrored and disconnected installs fail.

playbooks/tasks/collect_plugin_registries.yaml
eliorerz: This is how plugin-specific registry mirror entries (like `registry.redhat.io/lvms4` -> `quayHostname:8443/lvms4`) get into the container runtime config. Without it, image pulls for plugin operators fail in disconnected mode.

operators/openshift-pipelines-operator-rh/operator_update_pipeline.yaml
```diff
@@ -100,7 +100,7 @@
             script: |
                 #!/bin/bash
 
-                python ./approve_installplans.py "{{ storage_operators[blockStorageBackend] + operators }}" "$(params.dry-run)" > "$(results.status-report.path)"
+                python ./approve_installplans.py "{{ operators }}" "$(params.dry-run)" > "$(results.status-report.path)"
```
maorfr, line 103:
```suggestion
                python ./approve_installplans.py "{{ (_core_plugin_operators | default([])) + operators }}" "$(params.dry-run)" > "$(results.status-report.path)"
```
isn't this needed?

playbooks/tasks/configure_operators.yaml
```diff
@@ -2,14 +2,9 @@
...
-
     - name: Include configure_operator
       ansible.builtin.include_tasks: configure_operator.yaml
-      loop: "{{ storage_operators[blockStorageBackend] + operators }}"
+      loop: "{{ operators }}"
```
maorfr, line 7:
```suggestion
      loop: "{{ (_core_plugin_operators | default([])) + operators }}"
```
same here?

playbooks/tasks/configure_operator.yaml
maorfr: i think this is it: https://github.com/rh-ecosystem-edge/enclave/blob/main/bootstrap.sh#L175

playbooks/tasks/configure_operator.yaml
maorfr: all Subscriptions should be Manual, even in connected mode. except those in openshift-operators, which we currently have none.

i believe there should be no changes to this file.

playbooks/tasks/patch_mce_registries.yaml
maorfr: so instead of this file we need to template https://github.com/rh-ecosystem-edge/enclave/blob/main/operators/multicluster-engine/tasks.yaml similarly to registries.conf.j2?

playbooks/tasks/deploy_plugin.yaml
```diff
@@ -0,0 +1,179 @@
...
+    file: tasks/configure_operator.yaml
+    apply:
+      tags: operators
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+  loop: "{{ plugin.operators }}"
+  loop_control:
+    loop_var: operator
+  when: plugin.operators is defined
+  tags: operators
```
maorfr, line 169: i think this section should be removed?

templates/plugin-imageset.yaml.j2
```diff
@@ -0,0 +1,18 @@
+---
```
maorfr, line 1: i believe this will not be required given we mirror in imagesetconfiguration.yaml.j2?

schemas/plugin.yaml
```diff
@@ -0,0 +1,102 @@
...
+    description: Plugin type. foundation plugins deploy before core operators.
+  order:
+    type: integer
+    description: Deploy order among same-type plugins. Lower values deploy first.
+  mirror:
+    type: string
+    enum:
+      - core
+      - plugin
+      - none
```
maorfr, line 83: why do we need special handling for different operators? should we mirror everything through the same flow?

playbooks/tasks/pre_install_validate_plugins.yaml
```diff
@@ -0,0 +1,138 @@
+---
+- name: Add rendezvous node to inventory
+  ansible.builtin.add_host:
+    name: rendezvous
+    ansible_host: "{{ rendezvousIP }}"
+    ansible_user: core
+    ansible_ssh_common_args: "-o StrictHostKeyChecking=no -o BatchMode=yes -o ConnectTimeout=10"
```
agonzalezrh, line 7: Please add `-i {{ sshPubPath | replace('.pub','') }}`

plugins/odf/plugin.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - odf-csi-addons-operator
+      - rook-ceph-operator
+      - ocs-operator
+      - recipe
+      - mcg-operator
+      - odf-prometheus-operator
+      - ocs-client-operator
+      - cephcsi-operator
+      - odf-external-snapshotter-operator
+      - odr-cluster-operator
```
agonzalezrh, line 27: please rebase

playbooks/03-deploy.yaml
```diff
@@ -61,7 +61,24 @@
...
+        apply:
+          tags: hardware
+      tags: hardware
+
+    - name: Pre-install validate plugins
+      ansible.builtin.include_tasks:
+        file: tasks/pre_install_validate_plugins.yaml
+        apply:
+          tags: pre-install-validate
+      tags: pre-install-validate
```
agonzalezrh, line 74: Please edit bootstrap.sh 

`ansible-playbook playbooks/03-deploy.yaml -e@$global_vars -e@$certs_vars --tags hardware,pre-install-validate 2>&1 | tee -a ${log}`

playbooks/tasks/pre_install_validate_plugins.yaml
```diff
@@ -0,0 +1,138 @@
...
+# Ansible's retries/until mechanism, so we must wait for connectivity
+# explicitly before any delegated tasks.
+- name: Wait for rendezvous node to become reachable
+  ansible.builtin.wait_for_connection:
+    timeout: 300
+    sleep: 10
+  delegate_to: rendezvous
+
+- name: Get Assisted Service auth token from rendezvous node
+  ansible.builtin.shell: |
```
agonzalezrh, line 20: Could better replace this with ansible.builtin.slurp (https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/slurp_module.html)?

Then we avoid running commands directly, ping me if I can assist

agonzalezrh (changes requested):

PR #131, opened 2026-03-22
Avoid redundant mirrorring to internal quay on bootstrap

danielerez:
Added a configureQuayDisconnected flag in order to invoke mirroring to internal quay only on sync.
I.e. no need to mirror twice to the same registry during the bootstrap flow.

## Summary by CodeRabbit

* **New Features**
  * Added a new configuration flag to enable more granular control over Quay configuration in disconnected environments.

### DIFF
--- a/playbooks/06-day2.yaml
+++ b/playbooks/06-day2.yaml
@@ -49,7 +49,7 @@
       tags: openshift-pipelines
 
     - name: Configure Quay in disconnected environments
-      when: disconnected | default(true)
+      when: disconnected | default(true) and configureQuayDisconnected | default(false) | bool
       ansible.builtin.include_tasks:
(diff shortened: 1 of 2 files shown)


maorfr:
generally speaking, the day2 playbook is not intended to be executed in full.

danielerez:
> generally speaking, the day2 playbook is not intended to be executed in full.

we call it in nightly-e2e-disconnected: https://github.com/rh-ecosystem-edge/enclave/blob/d36b42b66f7d98edd766e581aff642008000ff83/.github/workflows/nightly-e2e-disconnected.yml#L181

should we skip mirror on day2 only in the e2e job?

maorfr:
i think we should skip calling the day2 playbook in e2e disconnected altogether.

PR #138, opened 2026-03-24
Combine pullSecrets using when instead of jinja templating

javipolo:
## Summary by CodeRabbit

* **Chores**
  * Refactored pull secret handling logic in operator and playbook deployment configurations by separating combined task definitions into distinct, explicitly-gated tasks for improved organization and clarity.


mlorenzofr (approved):
lgtm

PR #140, opened 2026-03-25
revert pull secret ambiguous type and add initial schema validation

maorfr:
reverts:
- #109
- #138
- #34 (updates)

main is currently broken: https://github.com/rh-ecosystem-edge/enclave/actions/runs/23526868007/job/68482070309

generally speaking, a variable should have a single type and that should be validated with schema validation. adding an initial schema validation in this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified pull-secret handling and ensured registry credential merges are consistently serialized for reliable cluster patching.

* **New Features**
  * Added variables schema validation to catch configuration issues earlier.

* **Documentation / Examples**
  * Example configuration updated to use object-style pullSecret with clearer replacement guidance.

* **Chores**
  * Validation tooling now reads example configuration from a file during checks.

### DIFF
--- a/config/global.example.yaml
+++ b/config/global.example.yaml
@@ -101,7 +101,7 @@ blockStorageBackend: lvms
 # Pull Secret and SSH Public Key Configuration
 # ============================================================================
 # Obtain from: https://console.redhat.com/openshift/install/pull-secret
-pullSecret: '{"auths":{"YOUR_PULL_SECRET"}}'
+pullSecret: {"auths":{}} # Replace with your Pull Secret
 
 # SSH public key path for cluster nodes
 sshPubPath: "YOUR_SSH_PUB_KEY_PATH"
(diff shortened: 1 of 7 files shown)


config/global.example.yaml
```diff
@@ -101,7 +101,7 @@ blockStorageBackend: lvms
 # Pull Secret and SSH Public Key Configuration
 # ============================================================================
 # Obtain from: https://console.redhat.com/openshift/install/pull-secret
-pullSecret: '{"auths":{"YOUR_PULL_SECRET"}}'
+pullSecret: {"auths":{}} # Replace with your Pull Secret
```
javipolo: Explicitly explain to NOT surround pull secret in single quotes. It happent to at least 2 people that we know already

javipolo (changes requested):

config/global.example.yaml
maorfr: done in https://github.com/rh-ecosystem-edge/enclave/pull/140/commits/8c5d3769d59d89f027c0179d35a3a4d5bc91f1b6

rporres:
got late for confirmation here, but LGTM since I was looking at it

config/global.example.yaml
rporres: that should be caught by schema check anyway, isn't it?

PR #153, opened 2026-03-26
[HOLD] Handle pre-existing Subscriptions in disconnected mode

eliorerz:
In disconnected mode, the default `CatalogSource` `redhat-operators` doesn't exist. Instead, you have a mirrored catalog, e.g. `cs-redhat-operator-index`. All our operator definitions set `operator.source: cs-redhat-operator-index`.
The problem is that some of the components create Subscriptions on their own. For example, the ingress operator automatically creates a `servicemeshoperator3` Subscription with `source: redhat-operators`. 

## Summary
- Add handling for pre-existing Subscriptions that point to non-existent CatalogSources in disconnected mode (e.g., `servicemeshoperator3` referencing `redhat-operators`)
- Creates a CatalogSource alias pointing to the mirrored catalog image
- Waits for InstallPlan, approves it, and waits for CSV to reach Succeeded state

## Context
This is part of a series of PRs splitting the `extract-storage-plugins` branch into independent, reviewable changes. This PR is a standalone improvement with no dependencies.

When operators are pre-installed (e.g., via OLM defaults), their Subscriptions may reference CatalogSources that don't exist in a disconnected environment. This block detects that situation and creates a mirrored CatalogSource alias so the operator can update properly.

## Test plan
- [ ] Verify `make validate` passes
- [ ] Deploy disconnected cluster with pre-existing operator Subscriptions
- [ ] Verify CatalogSource alias is created and operator updates successfully

## Summary by CodeRabbit

* **Chores**
  * Improved handling for pre-existing operator subscriptions that reference a different catalog source, enabling safe reconciliation and continuation of installs.
  * Automatic creation of catalog aliases when missing so installations can proceed despite source mismatches.
  * Added readiness checks, polling, and automatic approval of install plans and operator CSVs to ensure deployments reach a succeeded state.

### DIFF
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -56,6 +56,139 @@
     namespace: "{{ operator.namespace }}"
   register: r_sub_check
 
+# Handle pre-existing subscriptions that reference a non-existent CatalogSource.
+# For example, the ingress-operator creates a servicemeshoperator3 Subscription
+# with source: redhat-operators, which doesn't exist in disconnected mode.
+# The ingress-operator actively manages the subscription spec and overwrites any
+# changes, so instead of fighting it we create a CatalogSource alias that makes
+# the existing subscription resolve to the mirrored catalog.
+- name: "Handle pre-existing Subscription with wrong source (disconnected mode)"
+  when:
+    - r_sub_check.resources | length > 0
+    - disconnected | default(true) | bool
+    - r_sub_check.resources[0].spec.source != operator.source
+  block:
+    - name: "Check if CatalogSource '{{ r_sub_check.resources[0].spec.source }}' exists"
+      kubernetes.core.k8s_info:
+        api_version: operators.coreos.com/v1alpha1
+        kind: CatalogSource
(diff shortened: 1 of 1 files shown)


playbooks/tasks/configure_operator.yaml
```diff
@@ -56,6 +56,139 @@
...
+# The ingress-operator actively manages the subscription spec and overwrites any
+# changes, so instead of fighting it we create a CatalogSource alias that makes
+# the existing subscription resolve to the mirrored catalog.
+- name: "Handle pre-existing Subscription with wrong source (disconnected mode)"
+  when:
+    - r_sub_check.resources | length > 0
+    - disconnected | default(true) | bool
+    - r_sub_check.resources[0].spec.source != operator.source
+  block:
+    - name: "Check if CatalogSource '{{ r_sub_check.resources[0].spec.source }}' exists"
```
maorfr, line 71: i would set_fact to `r_sub_check.resources[0].spec.source` and give it a meaningful name

maorfr:
the discussion that led to holding this PR: we are not installing a subscription for servicemeshoperator3. for what operator do we encounter this problem than? because the initial condition is that a subscription is found.

PR #154, opened 2026-03-26
[HOLD] Add connected mode support for operator subscription approval

eliorerz:
## Summary
- Add `disconnected | default(true) | bool` condition to Manual subscription path so connected mode uses Automatic approval
- Simplify operator source to always use `{{ operator.source }}` instead of conditional logic
- Fix `startingCSV` string concatenation using Jinja2 `~` operator
- Use `k8s_retries`/`k8s_delay` variables consistently instead of hardcoded values

## Context
This is part of a series of PRs splitting the `extract-storage-plugins` branch into independent, reviewable changes. This PR is a standalone improvement with no dependencies.

Previously, Manual approval was used for all non-openshift-operators namespaces regardless of deployment mode. In connected mode, Automatic approval is simpler and avoids the need for InstallPlan approval automation.

## Test plan
- [ ] Verify `make validate` passes
- [ ] Deploy connected cluster — operators in custom namespaces use Automatic approval
- [ ] Deploy disconnected cluster — operators still use Manual approval with InstallPlan automation

### DIFF
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -59,7 +59,7 @@
 - name: "Ensure Subscription exists (Automatic)"
   when:
     - r_sub_check.resources | length == 0
-    - operator.namespace == "openshift-operators"
+    - operator.namespace == "openshift-operators" or not (disconnected | default(true) | bool)
   kubernetes.core.k8s:
     state: present
     definition:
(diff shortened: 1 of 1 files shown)


maorfr:
need a rebase?

maorfr:
why do we want to use Automatic in connected? the reason to use Manual is to be able to pin operators to a given version. if we use Automatic each operator will reach it's latest version, even if it's beyond the desired version.

PR #155, opened 2026-03-26
Add plugin schema, example plugin, and validation

eliorerz:
## Summary
- Define plugin contract via JSON Schema (`schemas/plugin.yaml`) specifying required fields, operator/registry definitions, plugin types (foundation/addon), and mirror modes
- Add example plugin (`plugins/example/`) as a template for new plugin development
- Add `validate_plugins.sh` script checking plugin structure, YAML validity, and required fields
- Add `validate-plugins` Makefile target and CI job in PR validation workflow
- Add `merge_group` trigger and `plugins/**` path filter to E2E deployment workflow

## Context
This is part of a series of PRs splitting the `extract-storage-plugins` branch into independent, reviewable changes. This PR establishes the plugin specification and validation — no runtime behavior changes. Subsequent PRs build on this foundation.

## Files
- `schemas/plugin.yaml` — JSON Schema for plugin descriptors
- `plugins/example/plugin.yaml` — Example addon plugin
- `plugins/example/tasks/{deploy,post-validate,pre-validate}.yaml` — Example lifecycle tasks
- `scripts/verification/validate_plugins.sh` — Plugin validation script
- `scripts/verification/validate.sh` — Added `plugins` validation function
- `Makefile` — Added `validate-plugins` target
- `.github/workflows/pr-validation.yml` — Added Plugin Validation job, `merge_group` trigger
- `.github/workflows/e2e-deployment.yml` — Added `merge_group` trigger, `plugins/**` path filter

## Test plan
- [ ] Run `make validate-plugins` — passes with example plugin
- [ ] Run `make validate` — includes plugin validation
- [ ] CI Plugin Validation job runs on PR

PR #163, opened 2026-03-26
Add plugin deployment framework

eliorerz:
## Summary
- Add plugin lifecycle runner (`deploy_plugin.yaml`) that executes deploy, pre-validate, and post-validate hooks
- Add foundation plugin auto-discovery (`deploy_foundation_plugins.yaml`)
- Add standalone playbook entry point (`deploy-plugin.yaml`)
- Add shell script `deploy_plugin.sh` for deploying individual plugins via SSH
- Add extra-args support to `deploy_phase.sh` for passing additional Ansible parameters
- Add per-plugin ImageSetConfiguration template (`plugin-imageset.yaml.j2`)
- Add `storage_plugin` and `enabled_plugins` to deployment defaults and schema
- Integrate foundation plugin deployment into `05-operators.yaml`

## Dependencies
- Merge after #155 (plugin schema + validation) - **already merged**

## Test plan
- [ ] `make validate` passes
- [ ] Plugin deployment framework loads without errors in Ansible check mode
- [ ] Foundation plugin auto-discovery finds plugins with `type: foundation`

## Summary by CodeRabbit

* **New Features**
  - Added explicit plugin discovery/deployment and a CLI to deploy named plugins.
  - Plugin mirroring support for air-gapped deployments.

* **Configuration Changes**
  - Introduced `storage_plugin` (lvms/odf) and `enabled_plugins` list.
  - Kept backward-compatible alias for `blockStorageBackend` to avoid breaking existing setups.

### DIFF
--- a/playbooks/05-operators.yaml
+++ b/playbooks/05-operators.yaml
@@ -4,7 +4,9 @@
 #
 # Performs:
 # - Disable default catalog sources (disconnected mode)
-# - Install storage operators (LVMS/ODF)
+# - Auto-discover and deploy foundation plugins (type: foundation)
+#   Storage plugins (LVMS/ODF) run here, BEFORE core operators,
+#   so that Quay can rely on the storage backend being present.
 # - Install general operators from defaults/operators.yaml
 #
 # Usage:
@@ -27,6 +29,29 @@
       ansible.builtin.debug:
         msg: "Installing operators in {{ 'disconnected' if (disconnected | default(true)) else 'connected' }} mode"
       tags: always
+
+    - name: Disable default operator catalog sources
+      when: disconnected | default(true) | bool
+      ansible.builtin.command: |
+        {{ workingDir }}/bin/oc patch OperatorHub cluster --type=merge -p='{"spec":{"disableAllDefaultSources":true}}'
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+      register: r_patch_operatorhub
+      changed_when: "'(no change)' not in r_patch_operatorhub.stdout"
+      tags:
+        - foundation-plugins
+        - operators
+
+    - name: Auto-discover and deploy foundation plugins
+      ansible.builtin.include_tasks:
+        file: tasks/deploy_foundation_plugins.yaml
+        apply:
+          tags:
+            - foundation-plugins
+            - operators
+      tags:
+        - foundation-plugins
+        - operators
 
     - name: Configure operators
       ansible.builtin.include_tasks:
--- a/config/global.example.yaml
+++ b/config/global.example.yaml
@@ -55,14 +55,22 @@ quayBackendRGWConfiguration:
 # ============================================================================
 # Storage Backend
 # ============================================================================
-# Storage backend for block devices for quay database and assisted installer
+# Storage plugin for block devices for quay database and assisted installer
+# Selects which storage plugin (plugins/lvms/ or plugins/odf/) to deploy
 # Options: lvms or odf
-# blockStorageBackend: odf
-blockStorageBackend: lvms
+# storage_plugin: odf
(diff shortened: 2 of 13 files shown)


maorfr:
let's focus this PR on day0 only. please remove the individual mirroring and instead find your way into the MCE configmap. individual mirroring is a significant paradigm shift and i'd like to isolate that to a distinct PR.

i also suggest to remove the entrypoints to the plugins from this PR (which is a day2 operation).

playbooks/05-operators.yaml
```diff
@@ -28,6 +30,29 @@
...
+        {{ workingDir }}/bin/oc patch OperatorHub cluster --type=merge -p='{"spec":{"disableAllDefaultSources":true}}'
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+      register: r_patch_operatorhub
+      changed_when: "'(no change)' not in r_patch_operatorhub.stdout"
+      tags:
+        - foundation-plugins
+        - operators
+
+    - name: Auto-discover and deploy foundation plugins
```
maorfr, line 45: i suggest to only discover here

playbooks/05-operators.yaml
```diff
@@ -28,6 +30,29 @@
...
+        file: tasks/deploy_foundation_plugins.yaml
+        apply:
+          tags:
+            - foundation-plugins
+            - operators
+      tags:
+        - foundation-plugins
+        - operators
+
     - name: Configure operators
```
maorfr, line 56: this will allow us to include the operator installation here

eliorerz:
> let's focus this PR on day0 only. please remove the individual mirroring and instead find your way into the MCE configmap. individual mirroring is a significant paradigm shift and i'd like to isolate that to a distinct PR.
> 
> i also suggest to remove the entrypoints to the plugins from this PR (which is a day2 operation).

I don't think that this is the right approach here, it will just add us more work now and later on. This is part of the design  and defining the plug-in lifecycle, also it will be included anyway.

maorfr:
we've seen in the past that installing operators individually introduces instability. day0 should mostly be the same flow. we need to remain 100% compatible with the current flows that are running in bootstrap and sync.

these are just some high level reasons to implement day0 first with minimal changes to the flow.

i believe that once we implement a clean day0 experience it will change the day2 implementation.

eliorerz:
> we've seen in the past that installing operators individually introduces instability. day0 should mostly be the same flow. we need to remain 100% compatible with the current flows that are running in bootstrap and sync.
> 
> these are just some high level reasons to implement day0 first with minimal changes to the flow.
> 
> i believe that once we implement a clean day0 experience it will change the day2 implementation.

  The whole point here is to change how things work today. If we need to change anything in bootstrap, let's do it. Assuming instability because previous attempts didn't work as expected isn't a reason to avoid the change this design was tested end-to-end in PR #130 with full connected and disconnected deployments passing CI and @agonzalezrh tests.

Also, The plugin framework doesn't change how operators are installed. The same `configure_operator.yaml` runs the same Subscription/InstallPlan/CSV flow. What changes is where the operator metadata lives, instead of `defaults/storage_operators.yaml`, it comes from `plugins/lvms/plugin.yaml`. The installation mechanism is identical.

Furthermore, This PR is the minimal set needed for the framework to function. Splitting it further would leave orphan infrastructure that can't be tested or that will require us to change the current design again.
That said, if the concern is specifically about the `patch_mce_registries.yaml` and `plugin-imageset.yaml.j2` (the mirroring-adjacent pieces), I can move those to the mirroring PR since no plugin uses them yet.

oourfali:
While I understand the concerns, there is a requirement to do it day2 as well. We don't have the luxury at the moment to keep discussing it further, so I'll take the decision here. We need to push this as it is today, and improve later as needed. We can't block this any longer.

PR #156, opened 2026-03-26
Split Ironic node configuration into boot and wait phases

eliorerz:
## Summary
- Split monolithic `configure_hardware_ironic_node.yaml` into two separate task files:
  - `configure_hardware_ironic_boot.yaml` — powers on nodes via Ironic/Redfish and initiates boot
  - `configure_hardware_ironic_wait.yaml` — waits for nodes to reach active state
- Update `03-deploy.yaml` to use boot + wait pattern with a gap between them

## Context
This is part of a series of PRs splitting the `extract-storage-plugins` branch into independent, reviewable changes. This PR is a standalone refactor with no dependencies.

The split creates a gap between boot and wait where pre-install validation can later be inserted (by a subsequent PR). This PR itself is functionally identical to the current behavior — boot then immediately wait.

## Files
- `playbooks/tasks/configure_hardware_ironic_boot.yaml` — NEW (extracted boot logic)
- `playbooks/tasks/configure_hardware_ironic_wait.yaml` — NEW (extracted wait logic)
- `playbooks/03-deploy.yaml` — Updated to use boot + wait pattern

## Test plan
- [ ] Verify `make validate` passes
- [ ] Deploy cluster — nodes boot and reach active state identically to current behavior

## Summary by CodeRabbit

* **Refactor**
  * Hardware provisioning split: boot initiation is now a separate per-host step, followed by a distinct per-host wait phase that ensures each node reaches active state before continuing.
* **Bug Fixes**
  * Deployment failures are now detected and reported in the dedicated wait phase; the boot step only initiates deployment and no longer blocks on node state.

PR #164, opened 2026-03-26
Add pre-install validation gate between Ironic boot and wait

eliorerz:
## Summary
- Add `pre_install_validate_plugins.yaml` that runs plugin pre-install validation hooks between Ironic boot and wait phases
- Add `run_pre_install_validate.yaml` for executing validation on the rendezvous node
- Insert pre-install validation step in `03-deploy.yaml` between boot and wait
- Add example pre-install-validate hook to the example plugin
- Add `pre-install-validate` tag to `bootstrap.sh`

## Dependencies
- Merge after #155 (plugin schema) - **already merged**
- Merge after #156 (Ironic boot/wait split) - provides the gap between boot and wait phases
-  Merge after #163 (Add plugin framework) - provides the plugin pre-validation step

## Test plan
- [x] `make validate` passes
- [x] Pre-install validation runs between boot and wait phases

## Summary by CodeRabbit

* **New Features**
  * Adds a pre-install validation phase that runs before hosts reach active state.
  * Hardware acquisition now triggers the pre-install validation step during deployment.
  * Automatic discovery and per-plugin pre-install checks are executed as part of deployment.
  * Example plugin logs discovered host count during pre-install validation.
  * Optional gating of automatic cluster start via an installation gate is supported.

### DIFF
--- a/playbooks/tasks/pre_install_validate_plugins.yaml
+++ b/playbooks/tasks/pre_install_validate_plugins.yaml
@@ -1,0 +1,156 @@
+---
+- name: Add rendezvous node to inventory
+  ansible.builtin.add_host:
+    name: rendezvous
+    ansible_host: "{{ rendezvousIP }}"
+    ansible_user: core
+    ansible_ssh_private_key_file: "{{ sshPubPath | replace('.pub', '') }}"
+    ansible_ssh_common_args: "-o StrictHostKeyChecking=no -o BatchMode=yes -o ConnectTimeout=10"
+
+# Nodes were just booted via Ironic and may not have SSH ready yet.
+# delegate_to connection failures (UNREACHABLE) are fatal and bypass
+# Ansible's retries/until mechanism, so we must wait for connectivity
+# explicitly before any delegated tasks.
+- name: Wait for rendezvous node to become reachable
+  ansible.builtin.wait_for_connection:
+    timeout: 300
+    sleep: 10
+  delegate_to: rendezvous
+
+- name: Read Assisted Service env file from rendezvous node
+  ansible.builtin.slurp:
+    src: /etc/assisted/rendezvous-host.env
+  delegate_to: rendezvous
+  register: r_rendezvous_env
+  retries: 30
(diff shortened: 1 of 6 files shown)


maorfr (approved):

playbooks/tasks/pre_install_validate_plugins.yaml
```diff
@@ -0,0 +1,145 @@
...
+    apply:
+      tags: pre-install-validate
+  loop: "{{ _pre_install_plugin_paths | default([]) }}"
+  loop_control:
+    loop_var: plugin_path
+  when: _pre_install_plugin_paths | default([]) | length > 0
+
+- name: Release installation gate on rendezvous node
+  ansible.builtin.file:
+    path: /var/tmp/proceed-with-installation
```
rporres: Apologies for the newbie question, but why is this a gate? What's waiting for this file to move on?

eliorerz: This is not a newbie question, you actually found a bug :)
In my whole PR I basically created a gate service that hold the installation so we could run validation. This service was waiting for that file to initiate the installation but I was suggested a better approach to handle this so this is not necessary anymore

eliorerz: Well, that approach didn't worked so we are back again to the gate service. see
https://github.com/rh-ecosystem-edge/enclave/blob/7a3e5aeed49883e5177d87164a55e515d99c6ad5/playbooks/tasks/configure_ocp_abi.yaml#L98-L109

eliorerz: So in ABI, `start-cluster-installation.service` is responsible for both creating the cluster in the Assisted Service and triggering installation. Unlike `dev-scripts`, which has a separate `create-cluster-and-infraenv.service` for cluster creation, ABI bundles everything into `start-cluster-installation.service`. By disabling it,
we prevented the cluster from being creating in the first place.

PR #168, opened 2026-03-27
Add CI image build workflow

rporres:
Build and push the CI image to quay.io/edge-infrastructure/enclave-lab-ci
on every merge to main via a dedicated ci-image.yml workflow.

In PR and merge queue runs, both pr-validation.yml and build-push-tarball.yml
resolve the CI image dynamically via a shared reusable workflow
(resolve-ci-image.yml): if Dockerfile.ci, ansible_collections.txt, or
ansible_pip_requirements.txt changed, build and push a transient image tagged
sha-<github.sha> with a 7-day Quay expiry label and use it for all validation
jobs; otherwise fall back to :latest. This eliminates the need for a second PR
after merging CI image changes, while keeping the common path free of build
overhead.

Update Dockerfile.ci to install Ansible Galaxy collections from
ansible_collections.txt and Python dependencies from ansible_pip_requirements.txt,
replacing the removed requirements.yml.

Fix the image namespace from quay.io/eerez/ to quay.io/edge-infrastructure/
across all workflows. Add CI_IMAGE_LABELS support to the Makefile
build-ci-image target to allow passing quay.expires-after at build time.

Switch base image from UBI9 to UBI10 (Python 3.12) to match the RHEL10
landing zone environment. Update EPEL to epel-release-latest-10.

Also remove the duplicate resolv.conf write at the end of the Dockerfile:
Docker/Podman always mount their own /etc/resolv.conf at container start,
overwriting anything set in the image. The first block at the top of the
Dockerfile is sufficient to ensure DNS works during the build.

## Summary by CodeRabbit

* **New Features**
  * Added an automated CI image build-and-publish workflow (manual and main-triggered) and a PR-aware image resolution step so CI jobs use the appropriate image.
  * CI image build now publishes both latest and SHA-tagged images for clearer versioning.

* **Chores**
  * Updated default CI image repository.
  * Pinned the base image to a digest for improved image reproducibility and security.

### DIFF
--- a/.github/workflows/Dockerfile.ci
+++ b/.github/workflows/Dockerfile.ci
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi
+FROM registry.access.redhat.com/ubi10/ubi@sha256:1b616c4a90d6444b394d5c8f4bd9e15a394d95dd628925d0ec80c257fdc5099c
 
 # Configure DNS
 RUN echo "nameserver 8.8.8.8" > /etc/resolv.conf && \
@@ -6,11 +6,10 @@ RUN echo "nameserver 8.8.8.8" > /etc/resolv.conf && \
 
 # Install EPEL for additional packages like shellcheck
 RUN dnf install -y \
-    https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm && \
+    https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm && \
     dnf clean all
 
 # Install system dependencies
-# UBI9 comes with curl-minimal, replace it with full curl
 RUN dnf install -y --allowerasing \
     ca-certificates \
     curl \
@@ -32,14 +31,17 @@ RUN dnf config-manager --add-repo https://download.docker.com/linux/rhel/docker-
 RUN pip3 install --no-cache-dir \
     yamllint \
     ansible-core \
-    ansible-lint \
-    jsonschema
+    ansible-lint
+
+# Install Python dependencies (kubernetes client, etc.)
+COPY ansible_pip_requirements.txt /tmp/ansible_pip_requirements.txt
+RUN pip3 install --no-cache-dir -r /tmp/ansible_pip_requirements.txt && \
+    rm /tmp/ansible_pip_requirements.txt
 
 # Pre-install Ansible Galaxy collections
-# Copy requirements.yml from repository root
-COPY requirements.yml /tmp/requirements.yml
-RUN ansible-galaxy collection install -r /tmp/requirements.yml && \
-    rm /tmp/requirements.yml
+COPY ansible_collections.txt /tmp/ansible_collections.txt
+RUN ansible-galaxy collection install -r /tmp/ansible_collections.txt && \
+    rm /tmp/ansible_collections.txt
 
 # Symlink collections to /usr/local for system-wide access
 RUN mkdir -p /usr/local/share/ansible/collections && \
@@ -50,9 +52,5 @@ ENV ANSIBLE_COLLECTIONS_PATH=/usr/local/share/ansible/collections:/root/.ansible
 
 # Set working directory
 WORKDIR /workspace
-
-# Configure DNS (in case it's needed at runtime)
-RUN echo "nameserver 8.8.8.8" > /etc/resolv.conf && \
-    echo "nameserver 8.8.4.4" >> /etc/resolv.conf
 
 CMD ["/bin/bash"]
--- a/.github/workflows/ci-image.yml
+++ b/.github/workflows/ci-image.yml
@@ -1,0 +1,39 @@
+name: CI Image Build
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-image-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ci-image:
+    name: CI Image Build
+    runs-on: [self-hosted, pr-validation]
+    timeout-minutes: 30
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build CI image
+        run: make -f Makefile.ci build-ci-image
+
+      - name: Test CI image
+        run: make -f Makefile.ci test-ci-image
+
+      - name: Push CI image
+        env:
+          QUAY_USER: ${{ secrets.QUAY_USER }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        run: make -f Makefile.ci push-ci-image
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -16,12 +16,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Job 0: Resolve CI image — build SHA-tagged image if Dockerfile.ci changed
+  resolve-image:
+    uses: ./.github/workflows/resolve-ci-image.yml
+    secrets: inherit
+
   # Job 1: Shell script validation
   shellcheck:
     name: Shell Script Validation
+    needs: resolve-image
     runs-on: [self-hosted, pr-validation]
     container:
-      image: quay.io/eerez/enclave-lab-ci:latest
+      image: ${{ needs.resolve-image.outputs.image }}
       options: --user root
     timeout-minutes: 5
     defaults:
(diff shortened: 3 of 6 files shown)


Makefile
```diff
@@ -159,7 +159,7 @@ validate-plugins:
 	@./scripts/verification/validate.sh plugins
 
 # CI Image targets
-CI_IMAGE_NAME ?= quay.io/eerez/enclave-lab-ci
+CI_IMAGE_NAME ?= quay.io/edge-infrastructure/enclave-lab-ci
```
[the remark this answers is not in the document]
rporres: I'll change it in a subsequent PR, when the build workflow pushes the image.

eliorerz: The Makefile changes `CI_IMAGE_NAME` to `quay.io/edge-infrastructure/enclave-lab-ci`, but pr-validation.yml and build-push-tarball.yml still reference `quay.io/eerez/enclave-lab-ci:latest`.

eliorerz (commented):
Nice work, I have few comments regarding this:

1.  If a PR adds a new tool to `Dockerfile.ci` and also uses that tool in a validation script, the PR will fail - the CI image is only pushed after merge. So that approach might not work here
2.  `/test ci-image` uses `rerun_pr_workflow`: This only works if the PR already has a `ci-image.yml` run (i.e., Dockerfile.ci or requirements.yml changed). If neither changed, the detect-changes job skips the build, and `/test ci-image` might fail.

Makefile
rporres: yes, I've done that since we don't have `quay.io/edge-infrastructure/enclave-lab-ci` yet until this PR has been merged.

rporres:
> If a PR adds a new tool to Dockerfile.ci and also uses that tool in a validation script, the PR will fail - the CI image is only pushed after merge. So that approach might not work here

Indeed. We would need to push from the PR too and establish dependencies between workflows. I don't really like it.

We can use an image with the sha from the workflows. That will signal that you cannot change the image and use it in the same run. That will properly set expectations.

rporres:
> /test ci-image uses rerun_pr_workflow: This only works if the PR already has a ci-image.yml run (i.e., Dockerfile.ci or requirements.yml changed). If neither changed, the detect-changes job skips the build, and /test ci-image might fail.

Hmmm, that is not a great idea. I'll remove then `/test ci-image`

Makefile
rporres: a complete new approach in d71f9a62f99cf1f57e74de90a0ec6e723a2aa41c

rporres:
a new approach in 0fa75a6e49b2bb90c9d5b393ccc60a067a568f97 so that we can use the new image in the same PR if needed, see commit message

.github/workflows/pr-validation.yml
```diff
@@ -16,12 +16,61 @@ concurrency:
...
+        run: make push-ci-image CI_IMAGE_TAG=sha-${{ github.sha }}
+
+      - name: Set image output
+        id: resolve
+        run: |
+          if [[ "${{ steps.filter.outputs.changed }}" == "true" ]]; then
+            echo "image=quay.io/edge-infrastructure/enclave-lab-ci:sha-${{ github.sha }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "image=quay.io/edge-infrastructure/enclave-lab-ci:latest" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
rporres: This finding is wrong for our use case — it would completely break the feature.

  CodeRabbit's suggestion adds && (github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch') to the build/test/push conditions. On a PR, github.ref is refs/pull/<n>/merge, not refs/heads/main, so the build steps would never run in a PR. The resolve-image job would always fall back to :latest — exactly the two-PR problem we set out to solve.

  The underlying security concern is: a PR author could modify the Makefile to push arbitrary content to the registry. That's a real concern for public repos with external contributors. But it doesn't apply here:

  - This is an internal project with a self-hosted runner
  - QUAY_USER/QUAY_TOKEN secrets are not available to forks — an external contributor's PR cannot push anything
  - The team already reviewed and accepted this trade-off earlier in the design discussion

  Verdict: do not apply. The finding is correct in principle for a public repo, but wrong for this project's context and would break the feature entirely.

.github/workflows/pr-validation.yml
```diff
@@ -16,12 +16,61 @@ concurrency:
...
+    outputs:
+      image: ${{ steps.resolve.outputs.image }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Check for CI image file changes
+        id: filter
+        uses: dorny/paths-filter@v3
```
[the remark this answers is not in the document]
rporres: since this pattern is widely used, let's tackle this in a subsequent PR, I don't want to touch unrelated things here

.github/workflows/resolve-ci-image.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: make push-ci-image CI_IMAGE_TAG=sha-${{ github.sha }}
+
+      - name: Set image output
+        id: resolve
+        run: |
+          if [[ "${{ steps.filter.outputs.changed }}" == "true" ]]; then
+            echo "image=quay.io/edge-infrastructure/enclave-lab-ci:sha-${{ github.sha }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "image=quay.io/edge-infrastructure/enclave-lab-ci:latest" >> "$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
rporres: fixed

.github/workflows/resolve-ci-image.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/checkout@v4
+
+      - name: Check for CI image file changes
+        id: filter
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            changed:
+              - '.github/workflows/Dockerfile.ci'
+              - 'ansible_collections.txt'
```
eliorerz, line 40: Don't we also need `ansible_pip_requirements.txt`?
```suggestion
          filters: |
            changed:
              - '.github/workflows/Dockerfile.ci'
              - 'ansible_collections.txt'
              - 'ansible_pip_requirements.txt'
              
```

eliorerz, line 40:
```suggestion
          filters: |
            changed:
              - '.github/workflows/Dockerfile.ci'
              - 'ansible_collections.txt'
              - 'ansible_pip_requirements.txt'
              
```

.github/workflows/ci-image.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: make -f Makefile.ci build-ci-image
+
+      - name: Test CI image
+        run: make -f Makefile.ci test-ci-image
+
+      - name: Push CI image
+        env:
+          QUAY_USER: ${{ secrets.QUAY_USER }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+        run: make -f Makefile.ci push-ci-image
```
eliorerz, line 39: The push step runs unconditionally, is that intended? 
Maybe we will want to push only on main, e.g.
```
github.ref == 'refs/heads/main'
```

.github/workflows/Dockerfile.ci
```diff
@@ -32,14 +31,17 @@ RUN dnf config-manager --add-repo https://download.docker.com/linux/rhel/docker-
 RUN pip3 install --no-cache-dir \
     yamllint \
     ansible-core \
-    ansible-lint \
-    jsonschema
```
eliorerz, line 36: I think we still need that, no?

rporres: It is in the ansible pip requirements, so we don't need to explicitly add it here. See https://github.com/rh-ecosystem-edge/enclave/blob/main/ansible_pip_requirements.txt#L12

.github/workflows/ci-image.yml
rporres: hmmm, there's a condition at the beginning of this file so that it is only triggered on push to main and workflow_dispatch.

eliorerz: Yeah, but my concern here is regarding the `workflow_dispatch`. It can be triggered from any branch via the GitHub UI or API. If someone dispatches it from a feature branch, it'll push `:latest` with that branch's image. I think that we want to allow updating latest only from main (even if it was triggered manually from another branch). WDYH?

eliorerz (changes requested):

.github/workflows/ci-image.yml
rporres: I see what you mean now. Fixed in 049a4ac30f29bf118111379c0c4cf5338c2de231

eliorerz (approved):

rporres:
Merged. Unrelated transient artifact error in E2E Disconnected

PR #177, opened 2026-03-29
Extract OpenShift AI and Nvidia GPU into plugins

maorfr:
this PR extracts OpenShift AI and Nvidia GPU into plugins.

following the example in #170

depends on (and contains):
- #176
- #178
- #179

## Summary by CodeRabbit

* **New Features**
  * NVIDIA GPU plugin for GPU support
  * OpenShift AI plugin to enable AI/ML capabilities
  * Plugin-level options: catalog selection, operator install toggle, additional images, and blocked images

* **Changes**
  * Removed centralized model configuration in favor of plugin-driven deployment
  * Deployment workflow now invokes plugins separately; operator mirroring and registry mappings updated

* **Documentation**
  * Deployment and configuration docs updated to reflect plugin-based approach


eliorerz:
We need to hold this one until the plugins pre/post validation will be verified.
Also we need to set pre/post validations for lvms/ODF.
I added to the PR title the WIP prefix so we will not merged it by mistake

maorfr:
> We need to hold this one until the plugins pre/post validation will be verified. Also we need to set pre/post validations for lvms/ODF. I added to the PR title the WIP prefix so we will not merged it by mistake

i can try to look into that. feel free to give a hint on what i can do to help!

agonzalezrh (approved):

playbooks/tasks/deploy_plugin.yaml
```diff
@@ -165,7 +165,9 @@
   loop: "{{ plugin.operators }}"
   loop_control:
     loop_var: operator
-  when: plugin.operators is defined
+  when:
+    - plugin.operators is defined
+    - plugin.mirror | default('none') == 'core'
```
eliorerz: I think this changes the gates operator installation on `plugin.mirror == 'core'`, which means addon plugins with `mirror: plugin` will never have their operators installed via the plugin runner. Both `openshift-ai` and `nvidia-gpu` fall into this category.
Basically the `mirror` field controls when images are mirrored. It shouldn't also control whether operators are installed. These should be separate concerns.

eliorerz: If we do want to control the operator installation you can expand the current implementation and add a new field like `install_operators: false` in `plugin.yaml`. WDYT?

bootstrap.sh
```diff
@@ -176,7 +176,8 @@ echo -p "Catalog source ACM policy .." -n1 -s
 echo -e "\e[38;5;10m Done...\033[0m"; date
 
 echo -p "Model config.. " -n1 -s
-    ansible-playbook playbooks/06-day2.yaml -e@$global_vars -e@$certs_vars --tags model-config 2>&1 | tee -a ${log}
+    ansible-playbook playbooks/deploy-plugin.yaml -e@$global_vars -e plugin_name=openshift-ai 2>&1 | tee -a ${log}
+    ansible-playbook playbooks/deploy-plugin.yaml -e@$global_vars -e plugin_name=nvidia-gpu 2>&1 | tee -a ${log}
```
eliorerz, line 180: Not sure if it is here only for testing but this kind of overrides the `enabled_plugins` variable and always deploy both regardless of the configuration.

plugins/openshift-ai/plugin.yaml
```diff
@@ -0,0 +1,152 @@
...
+  - location: "registry.redhat.io/rhoai"
+    mirror: "rhoai"
+  - location: "registry.redhat.io/rhods"
+    mirror: "rhods"
+  - location: "registry.redhat.io/rhelai1"
+    mirror: "rhelai1"
+  - location: "registry.redhat.io/rhaiis"
+    mirror: "rhaiis"
+  - location: "registry.redhat.io/openshift-service-mesh"
+    mirror: "openshift-service-mesh"
```
eliorerz, line 59: What about these two?
```yaml
  - location: "registry.redhat.io/openshift-service-mesh-tech-preview"
    mirror: "openshift-service-mesh-tech-preview"
  - location: "registry.redhat.io/openshift-service-mesh-dev-preview-beta"
    mirror: "openshift-service-mesh-dev-preview-beta"
```

operators/multicluster-engine/tasks.yaml
```diff
@@ -188,78 +188,12 @@
...
-          [[registry.mirror]]
-            location = "registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/rhaiis"
-
-          [[registry]]
-            prefix = ""
-            location = "registry.redhat.io/rhods"
-          [[registry.mirror]]
-            location = "registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/rhods"
-
           [[registry]]
```
eliorerz, line 263: I think that if an operator deploys the base cluster, provisions some managed clusters, and then later adds OpenShift AI, those existing managed clusters won't have the registry entries needed to pull AI images. They would need manual registry patching or re-provisioning. WDYT?

eliorerz: Worth documenting as an operational requirement, and potentially adding a post installation command e.g.  `make update-managed-cluster-registries` target or equivalent. It can be a dependency or run as part the `make deploy-plugin` target.

plugins/nvidia-gpu/plugin.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: nvidia-gpu
+type: addon
+order: 100
```
eliorerz: Don't we want it to be installed after the openshift AI? Maybe it worth setting the order to something higher than 100.

eliorerz (changes requested):
I liked the extended plugin schema here :) 
A few things to address before we can be merge it:

1. I would also add a genuine CI coverage for that - maybe something similar to what I did [here](https://github.com/rh-ecosystem-edge/enclave/blob/ac77f76eb0a6964c9794d29b2663d99e064f67a4/.github/workflows/e2e-nvidia.yml) for e2e-nvidia and [here](https://github.com/rh-ecosystem-edge/enclave/blob/ac77f76eb0a6964c9794d29b2663d99e064f67a4/.github/workflows/e2e-openshift-ai.yml) for OpenShift AI.

2. Lets add a `/test` entry point so reviewers can trigger plugin-specific e2e runs from PR comments (e.g. `/test e2e-openshift-ai`, `/test e2e-nvidia`)

3. Missing makefile targets for deploying specific plugin - see the [POC](https://github.com/rh-ecosystem-edge/enclave/blob/415d4835b272e028b9a7d0f129104378c12c0000/Makefile#L90-L94) for more info.

 4. I think that the 15 commits should be squashed into logical units before merge (there are "undo change", "minor fix", "ci fixes" commits that should be folded in).

playbooks/tasks/deploy_plugin.yaml
maorfr: good idea

bootstrap.sh
maorfr: i've removed the plugins from `enabled_plugins` but will keep them here for gori compatibility (at least for now).

plugins/openshift-ai/plugin.yaml
maorfr: these belong to earlier versions of servicemesh which we are no longer mirroring

operators/multicluster-engine/tasks.yaml
maorfr: so our current approach is to keep these here. got it.

plugins/nvidia-gpu/plugin.yaml
maorfr: sounds good

bootstrap.sh
maorfr: removed in #192

PR #197, opened 2026-04-01
Fix Quay DNS to use ingress VIP instead of router pod IP

eliorerz:
## Summary

- Replace ephemeral router pod IP with stable ingress VIP for Quay registry `/etc/hosts` entry
- Remove the `Get router pod for Quay DNS` task that fetched a pod IP (changes on restart)
- Use `{{ ingressVIP }}` which is constant regardless of pod scheduling

The router pod IP is ephemeral — when a pod restarts it gets a new IP, leaving `/etc/hosts` stale and Quay unreachable. The ingress VIP is a virtual IP managed by MetalLB/keepalived that stays constant.

Fixes: gori-project/GoRI#852
Relates: gori-project/GoRI#834

## Refactor

* Simplified registry endpoint configuration. Removed the process that discovers and monitors infrastructure components during setup. The registry hostname is now directly assigned using a Virtual IP address, streamlining deployment and reducing setup dependencies.


agonzalezrh (approved):
lgtm

PR #204, opened 2026-04-02
Add containerized Ceph setup for ODF CI testing

eliorerz:
## Summary
- Add infrastructure scripts to deploy a single-node Ceph cluster (via cephadm/podman) on CI runners with loopback-backed OSDs for ODF E2E testing
- Deploy scripts inject ODF external config and Quay RGW backend config from CI secrets when `storage-plugin=odf`
- E2E workflow verifies Ceph health before ODF deployments
- Add `/test e2e-connected-odf` and `/test e2e-disconnected-odf` slash commands

## Summary by CodeRabbit

* **New Features**
  * Added slash-command variants to trigger ODF-backed E2E runs.
  * Automated single-node Ceph provisioning, remote setup, and pre-flight verification for CI.
  * New Make targets to setup/verify Ceph on demand (ODF-only).
* **Chores**
  * CI workflows optionally run a Landing-Zone Ceph setup when ODF is selected and report failures from the new step.
  * Deployment flow now merges environment- or Landing-Zone–provided ODF/Quay config into provisioning.
  * Added utilities to attach loopback storage and activate Ceph volumes on runners.
* **Documentation**
  * Added comprehensive ODF/Ceph CI setup and usage guide.


scripts/deployment/deploy_cluster.sh
```diff
@@ -191,6 +191,10 @@ if [ -n "${ENABLED_PLUGINS:-}" ]; then
     info "Enabled plugins: $ENABLED_PLUGINS"
 fi
 
+# Load ODF external config and Quay RGW config (from env vars or LZ files)
+source "${ENCLAVE_DIR}/scripts/lib/odf.sh"
+append_odf_extra_vars
```
rporres: should we load this only if we're in odf. I mean something like

```bash
  if [ "${STORAGE_PLUGIN:-}" = "odf" ]; then
    source "${ENCLAVE_DIR}/scripts/lib/odf.sh"
    append_odf_extra_vars
  fi
```

and the same for the rest of the places where we load odf vars

docs/ODF_CEPH_CI.md
```diff
@@ -0,0 +1,208 @@
...
+
+No GitHub secrets needed. Config is generated and consumed within the same CI run.
+
+The `ODF_EXTERNAL_CONFIG` and `QUAY_BACKEND_RGW_CONFIG` environment variables are still supported for backward compatibility -- env vars take precedence over LZ config files.
+
+## CI Workflow Integration
+
+### Runner Labels
+
+ODF runs use the same runner labels as LVMS: `[self-hosted, enclave-large]`. No special `odf` runner label is required since Ceph is deployed dynamically on the Landing Zone.
```
rporres: This got me confused as we foresee in the code that we can have `odf` runners:

```bash
${{ matrix.storage-plugin == 'odf' && fromJSON('["self-hosted", "enclave-large", "odf"]') || fromJSON('["self-hosted", "enclave-large"]') }}
```

rporres (commented):
a couple of nits, looks great otherwise

scripts/deployment/deploy_cluster.sh
eliorerz: Good catch. Wrapped all three callsites (deploy_cluster.sh, deploy_phase.sh, deploy_plugin.sh) in `if [ "${STORAGE_PLUGIN:-}" = "odf" ]` so we skip sourcing odf.sh and the SSH checks for config files when running LVMS.

docs/ODF_CEPH_CI.md
eliorerz: You're right, the docs were wrong. The workflow does use `["self-hosted", "enclave-large", "odf"]` for ODF jobs. Updated the docs to match the odf label routes jobs to runners with enough disk space for the Ceph loopback OSDs.

rporres (approved):

PR #210, opened 2026-04-07
MGMT-23792: Remove rootfs download and serving

carbonin:
The rootfs URL has not been required in the AgentServiceConfig OSImages list in a long time. This is because the rootfs is always served from the image service full ISO content directly rather than requiring a separate stored artifact.

Supplying that URL was the only reason this image was downloaded and served so that can all just be removed. This saves time, complexity, and space (especially if many versions will eventually be configured).

Resolves https://redhat.atlassian.net/browse/MGMT-23792

For reference, the requirement for rootFSURL was removed several years ago in https://github.com/openshift/assisted-image-service/pull/83

## Release Notes

* **Refactor**
  * Removed support for RHCOS rootfs images; deployments now exclusively use RHCOS ISO images.
  * Consolidated image download processes for improved efficiency.

* **Documentation**
  * Updated configuration reference and deployment guides to reflect ISO-only image support.

* **Chores**
  * Removed deprecated rootfs image references from configuration schemas and service configurations.

### DIFF
--- a/defaults/content_images.yaml
+++ b/defaults/content_images.yaml
@@ -1,8 +1,5 @@
 ---
 content_images:
-  imgs:
-    - url: "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.20/4.20.0/rhcos-4.20.0-x86_64-live-rootfs.x86_64.img"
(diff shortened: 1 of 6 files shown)


carbonin:
Closing in favor of https://github.com/rh-ecosystem-edge/enclave/pull/211

PR #214, opened 2026-04-08
Add HTTPS support for Ironic vmedia server with CI trust chain

rporres:
Introduce end-to-end HTTPS serving for Ironic virtual media ISOs,
including a CA-based trust chain for CI and support for publicly
trusted certificates (e.g. Let's Encrypt).

HTTPS vmedia support (Ansible)
- Derive ironic_https_configured from ironicHTTPSCertificate/Key.
- Mount vmedia TLS cert/key as podman secrets into both the httpd and
  ironic containers; set IRONIC_EXTERNAL_IP and VMEDIA_TLS_PORT so
  Ironic generates https:// virtual media URLs on port 6183.
- Replace the temporary Python HTTP server with an /iso bind-mount
  into the Ironic container, using a file:// URL for boot_iso with
  OS_CONDUCTOR__FILE_URL_ALLOWED_PATHS=/iso.
- Health-check the HTTPS vmedia port with wait_for wrapped in a
  block/rescue: on timeout, capture podman logs httpd and re-fail
  with a descriptive message.
- Add pre-cleanup of stale Ironic pods before ephemeral deployment:
  stop Phase 7 persistent systemd services, remove root-owned and
  rootless pods. Prevents port 6180 conflicts when redeploying.
- Align quadlet templates (metal3-ironic-api.container.j2,
  metal3-httpd.container.j2) with ephemeral HTTPS/file-ISO changes
  for the persistent Phase 7 path.

Ironic hostname support
- Add optional lzBmcHostname variable so that publicly trusted
  certificates (DNS SANs only) can be used. When set, IRONIC_EXTERNAL_IP
  uses the hostname instead of lzBmcIP, producing vmedia URLs of the
  form https://<hostname>:6183/... that BMCs resolve via DNS.
  PROVISIONING_IP (Apache socket binding) remains lzBmcIP.
- Cert SAN validation in validations.sh checks against lzBmcHostname
  when set, falling back to lzBmcIP for the IP-SAN flow.

CA trust chain for CI
- generate_ironic_ca.sh: generate a per-run CA; write key+cert to
  ironic-ca/ and copy into the sushy-tools working directory.
- generate_ironic_cert.sh: after the Landing Zone is up, read
  lzBmcIP/lzBmcHostname via SSH, sign a server CSR with the CA,
  export cert+key+CA to GITHUB_ENV. Emits DNS+IP SAN when hostname
  is configured.
- start_sushy_tools.sh: install the CA into the container trust
  store via update-ca-trust.
- deploy_phase.sh: inject ENCLAVE_IRONIC_CERT, ENCLAVE_IRONIC_KEY,
  and ENCLAVE_IRONIC_CA into phase_vars.yaml.

CI workflow
- e2e-deployment.yml: enable HTTPS in e2e-connected; e2e-disconnected
  exercises the HTTP-only code path (port 6180, no certs injected).
- Makefile.ci: wrap generate_ironic_ca.sh in an ENCLAVE_IRONIC_HTTPS
  check so the disconnected run skips CA generation.

Schema and validation
- Add ironicHTTPSCertificate, ironicHTTPSKey, ironicHTTPSCACertificate,
  and lzBmcHostname to schemas/variables.yaml and definitions.yaml.
- Add schema test fixtures for valid and invalid HTTPS configurations.

https://redhat.atlassian.net/browse/MGMT-22877

## Summary by CodeRabbit

* **New Features**
  * Optional HTTPS support for serving boot ISOs/vmedia with configurable certificate, key, and CA.

* **Chores**
  * Automatic CA and short-lived server certificate generation and provisioning for vmedia HTTPS.
  * Hosting now supports HTTP or HTTPS and will restart when protocol, certificate, or key changes.
  * Deployment validates that certificate and key are provided together and installs the CA into local trust stores when present.
  * CI deployment flows add CA/cert generation steps and propagate HTTPS flags into deployments.

* **Tests/Validation**
  * Optional validation step now checks provided Ironic HTTPS certificates (skips if not provided).

### DIFF
--- a/config/certificates.example.yaml
+++ b/config/certificates.example.yaml
@@ -47,3 +47,18 @@ sslCACertificate: |
   -----BEGIN CERTIFICATE-----
   YOUR_ROOT_CA_CERTIFICATE
   -----END CERTIFICATE-----
+
+# Ironic HTTP Server TLS Certificate (optional)
+# Required when serving boot ISOs over HTTPS to the BMC.
+# The SAN on this certificate must cover lzBmcHostname (DNS SAN, for Let's Encrypt certs)
+# or lzBmcIP (IP SAN, for private CA certs). Set lzBmcHostname in global.yaml when using
+# a publicly trusted certificate such as Let's Encrypt.
+ironicHTTPSCertificate: ""
+
+ironicHTTPSKey: ""
+
+# Ironic HTTP Server TLS CA Certificate (optional)
+# Provide this when the BMC emulator must trust a private CA to verify
+# the Ironic HTTPS certificate. Installs the CA into the BMC emulator
+# trust store. Leave empty when using a publicly trusted certificate.
+ironicHTTPSCACertificate: ""
--- a/config/certificates.example.yaml
+++ b/config/certificates.example.yaml
@@ -56,9 +56,3 @@ sslCACertificate: |
 ironicHTTPSCertificate: ""
 
 ironicHTTPSKey: ""
-
-# Ironic HTTP Server TLS CA Certificate (optional)
-# Provide this when the BMC emulator must trust a private CA to verify
-# the Ironic HTTPS certificate. Installs the CA into the BMC emulator
-# trust store. Leave empty when using a publicly trusted certificate.
-ironicHTTPSCACertificate: ""
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -235,7 +235,7 @@ lzBmcIP: 100.64.1.10
 
 #### `lzBmcHostname`
 
-**Description**: DNS hostname for the LZ BMC interface. Optional. When set, Ironic constructs HTTPS vmedia URLs using this hostname instead of `lzBmcIP`.
+**Description**: DNS hostname for the LZ BMC interface. Optional. Only used when `ironicHTTPSCertificate` and `ironicHTTPSKey` are also set — ignored otherwise. When set, Ironic constructs HTTPS vmedia URLs using this hostname instead of `lzBmcIP`.
 
 **Type**: String (DNS hostname)
 
@@ -247,9 +247,10 @@ lzBmcHostname: mirror.example.com
 ```
 
 **Notes**:
+- Has no effect unless `ironicHTTPSCertificate` and `ironicHTTPSKey` are configured
 - The hostname must resolve to `lzBmcIP` from the BMC network
 - When set, `ironicHTTPSCertificate` must have a DNS SAN matching this hostname
-- When not set, the IP-based flow is used and `ironicHTTPSCertificate` must have an IP SAN matching `lzBmcIP`
+- When not set with HTTPS, `ironicHTTPSCertificate` must have an IP SAN matching `lzBmcIP`
 - `PROVISIONING_IP` (used for Apache binding) always remains `lzBmcIP`; only the vmedia URL changes
 
 #### `defaultNtpServers`
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -1144,23 +1144,6 @@ ironicHTTPSKey: |
   <PEM-encoded private key>
 ```
 
-#### `ironicHTTPSCACertificate`
-
-**Description**: CA certificate to install into the BMC emulator trust
-store. Only needed when the Ironic HTTPS certificate is signed by a
-private CA that the BMC emulator does not trust by default.
-
-**Type**: String (PEM format)
-
-**Required**: No. Leave empty when using a publicly trusted certificate.
-
-**Example**:
-```yaml
-ironicHTTPSCACertificate: |
-  -----BEGIN CERTIFICATE-----
-  ... (CA certificate)
-  -----END CERTIFICATE-----
-```
 
 ## Operator Configuration
 
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -233,6 +233,25 @@ lzBmcIP: 100.64.1.10
 - ISO will be served at `http://{{ lzBmcIP }}/assisted/agent.x86_64.iso`
 - Ensure HTTP server (Apache/Nginx) is running on this host
 
+#### `lzBmcHostname`
+
+**Description**: DNS hostname for the LZ BMC interface. Optional. When set, Ironic constructs HTTPS vmedia URLs using this hostname instead of `lzBmcIP`.
+
+**Type**: String (DNS hostname)
+
+**Required**: No. Set this when using publicly trusted TLS certificates (e.g. Let's Encrypt) for the Ironic vmedia server. Let's Encrypt only issues DNS SANs, not IP SANs, so the certificate SAN must match a hostname rather than an IP address.
+
+**Example**:
+```yaml
+lzBmcHostname: mirror.example.com
+```
+
+**Notes**:
+- The hostname must resolve to `lzBmcIP` from the BMC network
+- When set, `ironicHTTPSCertificate` must have a DNS SAN matching this hostname
+- When not set, the IP-based flow is used and `ironicHTTPSCertificate` must have an IP SAN matching `lzBmcIP`
+- `PROVISIONING_IP` (used for Apache binding) always remains `lzBmcIP`; only the vmedia URL changes
+
 #### `defaultNtpServers`
 
 **Description**: Optional list of additional NTP server addresses for cluster nodes. When not set, the cluster uses its default NTP sources.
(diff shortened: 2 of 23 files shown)


scripts/deployment/generate_ironic_ca.sh
```diff
@@ -0,0 +1,51 @@
...
+
+sudo openssl req -new -x509 -nodes -days 365 \
+    -subj "/CN=enclave-ironic-iso-ca" \
+    -keyout "${SUSHY_DIR}/ca.key" \
+    -out "${SUSHY_DIR}/ca.crt" \
+    2>/dev/null
+
+# Make CA key readable by the current user so generate_ironic_cert.sh
+# can sign the server cert without requiring sudo
+sudo chmod 644 "${SUSHY_DIR}/ca.key"
```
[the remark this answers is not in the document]
rporres: fixed

scripts/deployment/generate_ironic_cert.sh
```diff
@@ -0,0 +1,129 @@
...
+    } >> "$GITHUB_ENV"
+    success "Ironic ISO server certificate exported to GITHUB_ENV"
+else
+    info "GITHUB_ENV not set - printing certificate to stdout (local use)"
+    echo "ENCLAVE_IRONIC_CERT:"
+    echo "$CERT_CONTENT"
+    echo ""
+    echo "ENCLAVE_IRONIC_KEY:"
+    echo "$KEY_CONTENT"
+fi
```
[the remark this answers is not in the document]
rporres: fixed

playbooks/tasks/configure_hardware_ironic_setup.yaml
```diff
@@ -1,50 +1,160 @@
...
+      or iso_server_check.rc | default(1) != 0
+      or iso_server_protocol_mismatch | default(false)
+  changed_when: true
+
+- name: Start temporary Python ISO HTTPS server
   ansible.builtin.shell: >
-    nohup python3 -m http.server {{ http_server_port }} --directory {{ workingDir + '/ocp-cluster/abi-iso/' }} > {{ http_server_logfile }} 2>&1 &
-    echo $! > {{ http_server_pidfile }}
-  when: not pidfile_stat.stat.exists or http_server_check.rc != 0
+    nohup python3 {{ workingDir }}/scripts/deployment/https_server.py
```
carbonin: I'm less concerned with fetching these images over https since the connection will always be local.

If we need more security we could bind this to just a local address (rather than one that's exposed to the rest of the environment) or use a socket or something.

This seems like a bit more to maintain than is really necessary for a temp server.

What do you think?

rporres: I'll work on removing the python server and serve directly with the ironic httpd server

config/certificates.example.yaml
```diff
@@ -47,3 +47,16 @@ sslCACertificate: |
   -----BEGIN CERTIFICATE-----
   YOUR_ROOT_CA_CERTIFICATE
   -----END CERTIFICATE-----
+
+# Ironic HTTP Server TLS Certificate (optional)
+# Required when serving boot ISOs over HTTPS to the BMC.
+# The SAN on this certificate must cover lzBmcIP.
+ironicHttpsCertificate: ""
```
carbonin: I don't know how closely we're following conventions for naming here, but for go fields at least you typically don't mix case in an abbreviation. So this would be `ironicHTTPSCertificate` (and similar for the others).

You can see this in this file at least with things like `sslCACertificate` (rather than `sslCaCertificate`)

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -35,6 +35,25 @@
...
 
+- name: Check if Ironic HTTPS is configured
+  ansible.builtin.set_fact:
+    ironic_https_configured: >-
+      {{
+        ironicHttpsCertificate is defined and ironicHttpsCertificate | length > 0
+        and ironicHttpsKey is defined and ironicHttpsKey | length > 0
+      }}
+
+- name: Assert Ironic HTTPS inputs are complete
```
carbonin, line 52: Couldn't this just check the fact we set in the previous task?

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -44,6 +63,51 @@
     force: false
   no_log: true
 
+- name: Set Ironic HTTPS CA certificate file path
+  when:
+    - ironic_https_configured | bool
+    - ironicHttpsCaCertificate is defined
+    - ironicHttpsCaCertificate | length > 0
```
carbonin: Similar here. Aren't these checked when setting `ironic_https_configured`?

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -173,9 +256,11 @@
       - ALL
     volume:
       - metal3-ironic-shared:/shared
+    secrets: "{{ httpd_secrets }}"
     env:
       IRONIC_LISTEN_PORT: "6385"
       HTTP_PORT: "6180"
+      VMEDIA_TLS_PORT: "{{ ironic_vmedia_https_port | default('6183') }}"
```
carbonin: Does this really need to be set explicitly? Also if it does, do we need a variable for it or can we just set it directly (like the other ports here)?

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -206,6 +300,32 @@
...
 
+- name: Verify HTTPS vmedia server accessibility
+  when: ironic_https_configured | bool
+  block:
+    - name: Wait for HTTPS vmedia port
+      ansible.builtin.wait_for:
+        host: "{{ lzBmcIP }}"
+        port: "{{ ironic_vmedia_https_port | default(6183) }}"
+        timeout: 300
+  rescue:
```
carbonin, line 152: This might be a bit overkill. Do we do similar log captures in other cases?

rporres: This has been extremely useful debugging cert issues. I'd definitely keep it.

playbooks/tasks/deploy_ironic.yaml
rporres: No, we check there `ironicHTTPSCertificate` whereas we're checking here `ironicHTTPSCACertificate` (with the new var name)

playbooks/tasks/deploy_ironic.yaml
rporres: death by overconfiguration options 😂 

removed in 83ad7a6a8f62dd2c369545fb05b1fb1c8bc947e5

playbooks/tasks/deploy_ironic.yaml
rporres: simplified in 1ad30faade8d51a6aa24bcc9fec06bbdf35dd756

config/certificates.example.yaml
rporres: you're totally right. Fixed in e089262ff6861208036ef2d94dfc31828f4faf72

config/certificates.example.yaml
```diff
@@ -47,3 +47,18 @@ sslCACertificate: |
...
+# a publicly trusted certificate such as Let's Encrypt.
+ironicHTTPSCertificate: ""
+
+ironicHTTPSKey: ""
+
+# Ironic HTTP Server TLS CA Certificate (optional)
+# Provide this when the BMC emulator must trust a private CA to verify
+# the Ironic HTTPS certificate. Installs the CA into the BMC emulator
+# trust store. Leave empty when using a publicly trusted certificate.
+ironicHTTPSCACertificate: ""
```
carbonin: Based on the comments the only difference between `ironicHTTPSCertificate` and `ironicHTTPSCACertificate` is for the BMC emulator? Strange that this would be in the public configuration reference. I would figure this would be somewhere more CI specific.

Also there's really no difference between these names. Can we give this one a more descriptive name if it's really only for when we're using sushy tools (or decide if we can actually use the other var)?

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -41,6 +41,30 @@
...
+      }}
+
+- name: Assert Ironic HTTPS inputs are complete
+  when: >-
+    (ironicHTTPSCertificate is defined and ironicHTTPSCertificate | length > 0)
+    or (ironicHTTPSKey is defined and ironicHTTPSKey | length > 0)
+  ansible.builtin.assert:
+    that:
+      - ironic_https_configured | bool
+    fail_msg: "Set both ironicHTTPSCertificate and ironicHTTPSKey to enable Ironic HTTPS."
```
carbonin, line 59: Isn't this exactly the same assertion that sets the fact in "Check if Ironic HTTPS is configured"?

If I'm reading this right we set `ironic_https_configured` when the cert and key are defined and non-zero length. Then here we error if the fact is not set when the cert and key are defined and non-zero length. Surely we can just remove this task.

playbooks/tasks/deploy_ironic.yaml
```diff
@@ -51,6 +75,53 @@
...
+
+- name: Write Ironic HTTPS CA certificate to temp file
+  become: "{{ metal3_persistent | default(false) | bool }}"
+  when:
+    - ironic_https_configured | bool
+    - ironic_https_ca_configured | bool
+  ansible.builtin.copy:
+    content: "{{ ironicHTTPSCACertificate }}\n"
+    dest: "{{ ironic_ca_certfile }}"
+    mode: "0644"
```
carbonin: These seem like they belong in CI automation rather than in the prod playbooks.

playbooks/tasks/deploy_ironic_containers.yaml
```diff
@@ -1,6 +1,30 @@
 ---
 # Ironic ephemeral container deployment (non-persistent path)
 
+- name: Stop persistent Ironic systemd services if running
```
carbonin: Why add this in this PR? Seems unrelated.

playbooks/tasks/deploy_ironic.yaml
carbonin: Related to https://github.com/rh-ecosystem-edge/enclave/pull/214#discussion_r3183280044

playbooks/tasks/deploy_ironic_containers.yaml
rporres: I will take out from this PR and consider adding it later. This has given me problems while testing and I added it, but certainly not related to this PR

config/certificates.example.yaml
rporres: I've kept this one for completeness, but it is not actually used for the moment in the ironic playbook.

I understand that BMCs will trust whatever cert provided via ironicHTTPSCertificate, so this can actually be misleading since we're not doing anything. I'll remove the variable.

playbooks/tasks/deploy_ironic.yaml
rporres: I'll remove all the `ironicHTTPSCACertificate` related code from the playbooks, since it belongs to our CI only.

rporres: done in c809c51e7fb6eeea2809317318ca96893778bf32

config/certificates.example.yaml
rporres: removed that property in c809c51e7fb6eeea2809317318ca96893778bf32

playbooks/tasks/deploy_ironic.yaml
rporres: hmmm, no, here we're trying to catch partial configurations, hence the `or` in the `when` clause.

playbooks/tasks/deploy_ironic_containers.yaml
rporres: removed in 5bf5f9fa3ed43ca2b6f01f1a95247c39d48381b6

playbooks/tasks/deploy_ironic.yaml
carbonin: Ah okay. 👍

docs/CONFIGURATION_REFERENCE.md
```diff
@@ -233,6 +233,25 @@ lzBmcIP: 100.64.1.10
 - ISO will be served at `http://{{ lzBmcIP }}/assisted/agent.x86_64.iso`
 - Ensure HTTP server (Apache/Nginx) is running on this host
 
+#### `lzBmcHostname`
+
+**Description**: DNS hostname for the LZ BMC interface. Optional. When set, Ironic constructs HTTPS vmedia URLs using this hostname instead of `lzBmcIP`.
```
carbonin: Should this only be used when https is also being used? What if someone sets a hostname here, but isn't using https with ironic?

rporres: It is actually only used when HTTPS is only used. It has no effect otherwise.

clarified in the documentation in 747377b9124f73227111079c1f39cf65c496551c

carbonin (approved):

PR #215, opened 2026-04-08
Make GPU passthrough opt-in in CI workflows

eurijon:
GPU passthrough now requires `ENCLAVE_ENABLE_GPU_PASSTHROUGH=true`, preventing PCI conflicts when multiple CI jobs share a GPU host.
Also adds a check to skip passthrough when the GPU is already held by another VM.
Moves pre-install plugin validation after Ironic wait so deploy failures are caught early, and reduces the
rendezvous wait timeout from 900s to 600s accordingly.

MGMT-23797

### DIFF
--- a/playbooks/tasks/pre_install_validate_plugins.yaml
+++ b/playbooks/tasks/pre_install_validate_plugins.yaml
@@ -13,7 +13,7 @@
 # explicitly before any delegated tasks.
 - name: Wait for rendezvous node to become reachable
   ansible.builtin.wait_for_connection:
-    timeout: 900
+    timeout: 600
     sleep: 10
   delegate_to: rendezvous
 
--- a/scripts/infrastructure/configure_gpu_passthrough.sh
+++ b/scripts/infrastructure/configure_gpu_passthrough.sh
@@ -4,7 +4,8 @@
 # This script auto-detects an NVIDIA GPU on the host and attaches it
 # to the first master VM via PCI passthrough. It is designed to be
 # called as part of `make environment` and exits silently (exit 0)
-# when no GPU or no IOMMU is available.
+# when ENCLAVE_ENABLE_GPU_PASSTHROUGH is not set to "true", no GPU
+# is detected, or no IOMMU is available.
 #
 # Prerequisites (manual, one-time on GPU host):
 #   1. BIOS: VT-d (Intel) or AMD-Vi must be enabled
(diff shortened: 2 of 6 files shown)


scripts/infrastructure/configure_gpu_passthrough.sh
```diff
@@ -65,14 +77,28 @@ BUS=$(echo "$GPU_PCI_ADDRESS" | cut -d: -f2)
...
+# will hold the GPU.
+# We check all VMs (not just running) because the GPU is attached via --config
+# before the VM starts, so parallel jobs would both see it as available.
+GPU_IN_USE_BY=$(sudo virsh list --all --name | while read -r vm; do
+    [ -z "$vm" ] && continue
+    if sudo virsh dumpxml "$vm" 2>/dev/null | grep -q "bus='0x${BUS}'.*slot='0x${SLOT}'.*function='0x${FUNC}'"; then
+        echo "$vm"
+        break
+    fi
+done)
```
rporres: nit: I'd collect first the vms to avoid issues with `set -o pipefail`. Something like this (untested)

```
  mapfile -t _all_vms < <(sudo virsh list --all --name)
  GPU_IN_USE_BY=
  for vm in "${_all_vms[@]}"; do
      [ -z "$vm" ] && continue
      if sudo virsh dumpxml "$vm" 2>/dev/null | grep -q "bus='0x${BUS}'.*slot='0x${SLOT}'.*function='0x${FUNC}'"; then
          GPU_IN_USE_BY="$vm"
          break
      fi
  done
```

playbooks/tasks/pre_install_validate_plugins.yaml
```diff
@@ -13,7 +13,7 @@
 # explicitly before any delegated tasks.
 - name: Wait for rendezvous node to become reachable
   ansible.builtin.wait_for_connection:
-    timeout: 900
+    timeout: 600
```
rporres, line 16: nit: you may want to fix the PR description to match this value.

rporres (approved):
couple of nits, looks fine overall

eurijon:
> couple of nits, looks fine overall

Thanks Rafa, let me address those issues!

scripts/infrastructure/configure_gpu_passthrough.sh
eurijon: Done!

PR #221, opened 2026-04-09
Make Metal3 stack persistent via quadlets (flag-controlled)

rporres:
Add `metal3_persistent` flag (default: false) to `deploy_ironic.yaml` and `deploy_bmo.yaml`. When false (phase 3), the existing ephemeral podman behaviour is preserved. When true (phase 7), quadlet files are deployed to `/etc/containers/systemd/` so systemd manages the Metal3 services and they survive an LZ reboot.

New quadlet templates:
- `metal3-ironic.pod.j2` (Network=host pod)
- `metal3-ironic-api.container.j2` (Pull=never, conditional SSL CA)
- `metal3-httpd.container.j2`
- `metal3-bmo.container.j2` (After=metal3-ironic-api.service)

New task files:
- `deploy_ironic_containers.yaml` – ephemeral path (extracted from deploy_ironic.yaml)
- `deploy_ironic_quadlet_units.yaml` – persistent path
- `deploy_bmo_container.yaml` – ephemeral path (extracted from deploy_bmo.yaml)
- `deploy_bmo_quadlet_unit.yaml` – persistent path
- `configure_metal3_teardown.yaml` – manual decommission helper

`07-configure-discovery.yaml` sets `metal3_persistent: true` before the Metal3 deployment tasks so phase 7 uses the quadlet path.

https://redhat.atlassian.net/browse/MGMT-22878

Signed-off-by: Rafa Porres Molina <rporresm@redhat.com>

Assisted-by: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Persistent Metal3 deployment via systemd/quadlet units with image pre-pull for faster, reliable starts.
  * New ordered teardown workflow to stop services, remove containers/pods, and optionally remove volumes and secrets.

* **Refactor**
  * Split flows into ephemeral vs persistent paths for Metal3 components; modularized tasks for clearer, maintainable deployments.

* **Security**
  * Hardened operator runtime: non‑root execution, dropped capabilities, and restricted secret mounts.


playbooks/tasks/deploy_ironic_quadlet_units.yaml
```diff
@@ -0,0 +1,100 @@
...
+  ansible.builtin.file:
+    path: /etc/containers/systemd
+    state: directory
+    mode: '0755'
+
+- name: Deploy metal3-ironic pod quadlet
+  become: true
+  ansible.builtin.template:
+    src: "{{ playbook_dir }}/templates/quadlets/metal3-ironic.pod.j2"
+    dest: /etc/containers/systemd/metal3-ironic.pod
```
carbonin, line 15: Are we ever expected to run multiple instances of this in a single LZ?
I recall someone talking about that at some point. Maybe it was just for CI though?

In either case this will be global, right?

playbooks/tasks/deploy_bmo_container.yaml
```diff
@@ -0,0 +1,52 @@
...
+      WATCH_NAMESPACE: infraenv
+      LIVE_ISO_FORCE_PERSISTENT_BOOT_DEVICE: Never
+      METAL3_AUTH_ROOT_DIR: /auth
+    command:
+      - /baremetal-operator
+      - --webhook-port=0
+      - --metrics-addr=0
+      - --health-addr=0
+      - --dev
+      - --enable-leader-election=false
```
carbonin, line 30: Is there any way we can deduplicate this?
It's unfortunate that we need to carry all this information here as well as in the quadlet template.

Is it possible to also run the quadlet for the first run then tear it down and run it again?

playbooks/tasks/deploy_ironic_quadlet_units.yaml
rporres: Not that I know of. I don't know why we would need to run multiple instance of this. Anyway it is not how it currently works, which is what I'm trying to keep.

Yes it will be global.

playbooks/tasks/deploy_bmo_container.yaml
rporres: I thought about that but I decided on focus on persistence in phase 7 and leave things in phase 3 much as they are now since playbooks are independent and it looks too much to add quadlets in phase 3.

That has the downside of some duplication but it looked to be right thing given the fact that we're tearing down the whole stack once the cluster has been installed

playbooks/tasks/configure_metal3_teardown.yaml
```diff
@@ -0,0 +1,110 @@
+---
+# Metal3 teardown tasks - decommission persistent Metal3 deployment
```
carbonin: Is this file used anywhere? I don't see it referenced elsewhere in this PR.

rporres: no, I just added it for completeness if needed. But it's true that it can be misleading.   Removed in b0940afc0bca8db6d09b628c428a7fdc32fa9748

carbonin (approved):

PR #228, opened 2026-04-13
revert operator update functionality and patch subscription to manual to avoid further upgrades

maorfr:
this PR removed the operator update functionality and changes the method in which we install operators.

we go back to installing with an Automatic install plan approval (prior to #16), and we eventually set the install plan approval to Manual to avoid upgrades when new content is mirrored.

we still only mirror operator versions up to the desired version, which means we end up with the same state of installed operators and their versions.

the reason for this change is to overcome intermittent installation failures resulting from operator dependencies, this time due to alternating order of approvals. the method reverted in this PR can not be relied, as this is not the first time we are making changes in it, specifically for the odf case (previously: #124)

## Summary by CodeRabbit

* **Refactor**
  * Removed the standalone operator update pipeline that automated install-plan approvals.
  * Integrated install-plan approval logic into the subscription configuration workflow.
  * Unified subscription creation and status verification into a single flow, removing the separate approval step.
  * Simplified operator deployment by eliminating the external approval helper and pipeline orchestration.


maorfr:
related to https://github.com/gori-project/GoRI/issues/873

rporres:
can you add details in the description on what you're exactly trying to fix?

pgodowski:
I suggest you put this on hold - the issue https://github.com/gori-project/GoRI/issues/873 turned out to be intermittent. Changing so late the entire install paradigm is too late and not acceptable on our side.

maorfr:
> I suggest you put this on hold - the issue [gori-project/GoRI#873](https://github.com/gori-project/GoRI/issues/873) turned out to be intermittent. Changing so late the entire install paradigm is too late and not acceptable on our side.

the reason for this change is to overcome intermittent installation failures resulting from operator dependencies, this time due to alternating order of approvals. the method reverted in this PR can not be relied, as this is not the first time we are making changes in it, specifically for the odf case (previously: #124)

rporres (approved):

pgodowski:
https://github.com/rh-ecosystem-edge/enclave/pull/124 was added month ago and we had time to validate. This rework removes entirely the manual install plan approach, not only after DCUT (we have to ship in 2 weeks, remember), but also to address intermittent issue. Every change on your side requires pre-validation on our environments (we test on bare metals), which might uncover some new issues, leading to the congestion in picking up the real high sev issues.

Put this fix into the future branch please, but not into the GA branch.

operators/openshift-pipelines-operator-rh/tasks.yaml
```diff
@@ -15,10 +15,6 @@
   retries: 120
   delay: 30
 
-- name: Create Operator Update Pipeline
```
pgodowski, line 18: please put it into the future release branch

pgodowski (changes requested):

PR #229, opened 2026-04-13
extract odf quay task resources to plugin defaults

maorfr:
extract the resources section of the quay components as required by odf plugin to the plugin file.

the defaults are later used to template the task and apply the required values.

this approach enables and signals a consumer to get involved in setting the values should they choose to, via the plugin file which is the interface to the functionality.

## Summary by CodeRabbit

* **Chores**
  * Updated default resource configuration for Postgres components with standardized CPU (1 limit / 500m request) and memory (4Gi limit / 2Gi request) allocations.
  * Migrated resource specifications to use templated configuration values for improved consistency and manageability across deployments.

### DIFF
--- a/plugins/odf/plugin.yaml
+++ b/plugins/odf/plugin.yaml
@@ -42,6 +42,25 @@ requires:
 defaults:
   odfDefaults:
     defaultStorageClass: true
+  components:
+    postgres:
+      overrides:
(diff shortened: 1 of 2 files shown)


maorfr:
related to https://github.com/gori-project/GoRI/issues/871

PR #230, opened 2026-04-13
Move test-fixtures to its own directory

rporres:
This will mark clearly what those files are used for. I'm not excluding it for tarball as scripts like `scripts/verification/validate.sh` use the fixtures. But location will mark clearly enough what those files are used for

scripts/verification/validate.sh

https://redhat.atlassian.net/browse/MGMT-23837

https://github.com/gori-project/GoRI/issues/874


rporres:
E2E error is unrelated

maorfr (approved):

PR #233, opened 2026-04-14
[MGMT-23811] enable build-push-tarball github action for 0-rc branch

maorfr:
part of https://redhat.atlassian.net/browse/MGMT-23811

this PR enables the github actions that are enabled for main.

while these changes have no effect in main, we want to have 0-rc as similar as possible to main. this is why we are adding 0-rc in main. this will allow us to have 0-rc cleanly track main while we get it started.

### DIFF
--- a/.github/workflows/build-push-tarball.yml
+++ b/.github/workflows/build-push-tarball.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - 0-rc
     tags:
       - '*'
   pull_request:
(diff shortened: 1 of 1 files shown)


maorfr:
@coderabbitai review

rporres:
I understand that "Exit if image exists" is made to avoid race conditions between pushes in the two branches. But I think the problem is that, the way it is written, it is going to be a no-op if the image exists too and it will also end trying to run all the steps in that case

I think you need to make the rest of the steps dependent, e.g. (untested)

```
  - name: Check if image exists
    id: check
    continue-on-error: true
    run: |
      oras manifest fetch quay.io/edge-infrastructure/enclave:${{ steps.meta.outputs.tag }} > /dev/null 2>&1

  - name: Build tarball
    if: steps.check.outcome == 'failure'
    run: |
      ...

  - name: Validate tarball
    if: steps.check.outcome == 'failure'
    run: |
      ...

  - name: Push tarball to Quay
    if: steps.check.outcome == 'failure'
    run: |
      ...
```

PR #244, opened 2026-04-14
do not apply updated mirror manifests to cluster if operators should not be installed

maorfr:
if operators should not be installed, they shouldn't be applied to IDMS/ITMS on the management cluster.

- https://github.com/rh-ecosystem-edge/enclave/blob/7204790fc9bfaf9f5aa1a501e6f622d415e41c1d/plugins/openshift-ai/plugin.yaml#L41
- https://github.com/rh-ecosystem-edge/enclave/blob/7204790fc9bfaf9f5aa1a501e6f622d415e41c1d/plugins/nvidia-gpu/plugin.yaml#L14


maorfr:
related to https://github.com/gori-project/GoRI/issues/876

agonzalezrh (approved):

PR #245, opened 2026-04-14
plugin Apply Quay mirrors and trust registry CA for image pulls

maorfr:
use existing implementation from #119

this can be a single location to gradually improve the required operations after mirroring new content that is required on the management cluster.

ref: https://github.com/rh-ecosystem-edge/enclave/blob/main/operators/quay-operator/quay_disconnected_mirrors.yaml


maorfr:
closing as it's not needed for now, but the idea still applies

PR #247, opened 2026-04-14
add oc-mirror from landing zone to quay-enterprise in deploy plugin

maorfr:
seems we are only mirroring to the landing zone mirror registry.

this PR adds another oc-mirror call to mirror content from landing zone mirror registry to quay-enterprise.

## Summary by CodeRabbit

* **Infrastructure**
  * Extended plugin image mirroring to support Quay Enterprise registry with specialized configuration for authentication, TLS verification, and dynamic optimization based on storage backend.
  * Improved deployment task organization through naming updates for clarity.

### DIFF
--- a/playbooks/tasks/deploy_plugin.yaml
+++ b/playbooks/tasks/deploy_plugin.yaml
@@ -124,7 +124,7 @@
     - plugin.operators is defined
   tags: mirror
 
-- name: Run oc-mirror for plugin
+- name: Run oc-mirror for plugin (Landing Zone)
   ansible.builtin.shell: |
     {{ workingDir }}/bin/oc-mirror --v2 \
       --log-level {{ ocMirrorLogLevel }} \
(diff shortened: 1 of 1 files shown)


playbooks/tasks/deploy_plugin.yaml
```diff
@@ -147,6 +147,31 @@
...
+      > {{ workingDir }}/logs/oc-mirror-plugin-{{ plugin_name }}.progress.quay.$(date +%s).log 2>&1
+  retries: 10
+  delay: 10
+  register: r_oc_mirror_quay
+  until: r_oc_mirror_quay is succeeded
+  when:
+    - plugin.mirror | default('none') == 'plugin'
+    - disconnected | default(true) | bool
+  tags: mirror
+
```
eliorerz, line 174: I think that this step is missing two things that the core Quay mirror (in `quay_disconnected.yaml`) does before running oc-mirror:

  1. **`registries.conf`**: The core Quay mirror generates `~/.config/containers/registries.conf` from `registries.conf.j2`, which tells oc-mirror to pull from the LZ registry (`quayHostname:8443`) instead of upstream (`registry.redhat.io`). Without it, oc-mirror will try to reach upstream registries, which will fail in a disconnected environment.
The template also doesn't include addon plugin registries (NVIDIA, OpenShift AI), only core plugin registries via `_core_plugin_registries`. So even if you generate the registries.conf, you'd need to add the plugin's registry entries to it.

2. **Imageset catalog rewrite**: The core Quay mirror copies the imageset to
     `imagesetconfiguration.internal.yaml` and rewrites `catalog: registry.redhat.io` ->
     `catalog: quayHostname:8443`. This task uses the original plugin imageset pointing to upstream catalogs.

eliorerz (changes requested):

maorfr:
replaced by #248

PR #253, opened 2026-04-15
MGMT-23910: Add --step and --non-interactive flags to bootstrap.sh

eliorerz:
## Summary
- Refactor `bootstrap.sh` to extract each deployment phase into a separate function (`step_setup`, `step_validate`, `step_download_content`, `step_build_cache`, `step_acquire_hardware`, `step_deploy`, `step_post_install`, `step_operators`, `step_day2`, `step_discovery`)
- Add `--step <name>` flag to run a single step instead of all steps
- Add `--non-interactive` flag to skip the "Press Enter" prompt
- Without flags, behavior is identical to before (backward compatible)

This is the first PR in a series to make CI run `bootstrap.sh` directly on the LZ, closing the gap between CI and user-facing flows.

## Test plan
- [ ] `./bootstrap.sh` without flags produces the same behavior as before
- [ ] `./bootstrap.sh --step validate --non-interactive` runs only validation
- [ ] `./bootstrap.sh --step download-content --non-interactive` runs only Phase 1a
- [ ] `./bootstrap.sh --step invalid-name` exits with a clear error
- [ ] `./bootstrap.sh --help` shows all available steps

## Summary by CodeRabbit

* **New Features**
  * Added a --step option to run an individual pipeline step by name.
  * Added a --non-interactive mode to suppress interactive start prompts.
  * Introduced a new "Partner Overlay" step in the pipeline.

* **Refactor**
  * Restructured the execution flow into modular, discrete steps with centralized dispatch logic.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -8,13 +8,30 @@ usage() {
     echo "Options:"
     echo "  --global-vars FILE    Path to global vars file (default: config/global.yaml)"
     echo "  --certs-vars FILE     Path to certificates vars file (default: config/certificates.yaml)"
+    echo "  --step STEP           Run a single step instead of all steps"
+    echo "  --non-interactive     Skip interactive prompts"
     echo "  -h, --help            Show this help message"
+    echo ""
+    echo "Available steps:"
+    echo "  setup               Configure environment (setup_env + setup_ansible)"
+    echo "  validate            Validate configuration (schema + validations)"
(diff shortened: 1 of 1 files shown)


eurijon:
What about adding Partner OverLay script execution in a separate step too? it would follow the overall step approach and would avoid running `./partner-install/start.sh` each time `bootstrap.sh` is called with a specific step.

maorfr (approved):

PR #254, opened 2026-04-15
MGMT-23910: Wire CI to run bootstrap.sh steps via SSH

eliorerz:
## Summary
- Add `scripts/deployment/deploy_bootstrap_step.sh` — SSH wrapper that runs `bootstrap.sh --step <name> --non-interactive` on the Landing Zone
- Add `deploy-bootstrap-*` Makefile.ci targets for each step (setup, validate, download-content, build-cache, acquire-hardware, deploy, post-install, operators, day2, discovery)
- Replace `deploy-cluster-*` steps in both e2e-connected and e2e-disconnected workflows with `deploy-bootstrap-*` steps
- Update failed-step detection maps in both jobs to use the new step IDs

**Depends on:** #253

Old `deploy-cluster-*` targets are kept for backward compatibility.

## Test plan
- [ ] CI runs each bootstrap step as a separate GitHub Actions step
- [ ] Per-step visibility is preserved in the GitHub UI
- [ ] Failed-step detection works with new step IDs
- [ ] Slack notifications show the correct failed step name
- [ ] Old `deploy-cluster-*` Makefile targets still work

## Summary by CodeRabbit

* **New Features**
  * CI now auto-generates self-signed certificates for the cluster API and apps wildcard for testing.

* **Chores**
  * CI deployment redesigned into a single ordered bootstrap pipeline with per-stage execution and clearer step/status reporting.
  * Runtime host/OS eligibility checks moved into bootstrap setup.
  * Ingress host list expanded to include an additional apps entry; discovery ordering tightened.


maorfr (approved):
LGTM

I suggest to keep the lingo in bootstrap to "disconneted" instead of adding "is_connected"

PR #261, opened 2026-04-16
fixes for plugins catalogsources - custom per-plugin catalogsource

maorfr:
this PR adds:
- plugin mirror to custom per-plugin catalogs
- catalgsource resources for each plugin with the custom per-plugin catalog name and image
- usage of the custom per-plugin catalogs in the plugins tasks (when deploying operators via acm policies)
- a policy dependency on a new policy that applies the catalogsource and verifies it is ready
- servicemeshoperator3 version and catalogsource fixes/hacks

related to:
- https://github.com/gori-project/GoRI/issues/883
- https://github.com/gori-project/GoRI/issues/884
- https://github.com/gori-project/GoRI/issues/885

## Summary by CodeRabbit

* **New Features**
  * Added policies to provision mirrored operator catalog sources for NVIDIA GPU and OpenShift AI and enforce their READY state before installation.
  * Made operator installation policies explicitly depend on the corresponding catalog-source policies and updated placement bindings to include them.

* **Chores**
  * Updated deployment tasks, subscription sources, and image set templates to target mirrored catalog names per plugin type.
  * Removed an outdated catalog-source entry for a previously referenced certified catalog.

### DIFF
--- a/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
+++ b/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
@@ -9,7 +9,7 @@ metadata:
   namespace: openshift-acm-policies
 spec:
   disabled: false
-  remediationAction: inform
+  remediationAction: enforce
   policy-templates:
   - objectDefinition:
       apiVersion: policy.open-cluster-management.io/v1
--- a/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
+++ b/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
@@ -31,7 +31,7 @@ spec:
               name: {{ mirror_rh_operator_catalog }}-openshift-ai
               namespace: openshift-marketplace
             spec:
-              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/{{ mirror_rh_operator_catalog }}-openshift-ai:v{{ short_version_dot }}
+              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/{{ mirror_rh_operator_catalog }}-openshift-ai:{{ rh_operator_catalog_version }}
               sourceType: grpc
         - complianceType: musthave
           metadataComplianceType: musthave
--- a/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
+++ b/plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
@@ -1,0 +1,62 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  annotations:
+    policy.open-cluster-management.io/categories: CM Configuration Management
+    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
+    policy.open-cluster-management.io/standards: NIST SP 800-53
+  name: model-catalogsources
+  namespace: openshift-acm-policies
+spec:
+  disabled: false
+  remediationAction: inform
+  policy-templates:
+  - objectDefinition:
+      apiVersion: policy.open-cluster-management.io/v1
+      kind: ConfigurationPolicy
+      metadata:
+        name: model-catalogsources
+      spec:
+        evaluationInterval:
+          compliant: 5m
+          noncompliant: 45s
+        object-templates:
+        # Create custom CatalogSource for the plugin
+        - complianceType: mustonlyhave
+          metadataComplianceType: musthave
+          objectDefinition:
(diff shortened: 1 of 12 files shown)


maorfr:
this PR combines #258, #259 and #260

plugins/openshift-ai/plugin.yaml
```diff
@@ -36,7 +36,7 @@ operators:
   - name: servicemeshoperator3
     version: 3.2.2
     channel: stable
-    init_version: 3.2.2
+    init_version: 3.1.0
```
rporres, line 39: what's this for?

maorfr: https://github.com/gori-project/GoRI/issues/883#issuecomment-4257692708

rporres: I see. Can you add a note in the PR description?

rporres (approved):
The approach looks solid 🙏

plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
```diff
@@ -0,0 +1,73 @@
...
+            apiVersion: operators.coreos.com/v1alpha1
+            kind: CatalogSource
+            metadata:
+              name: {{ mirror_rh_operator_catalog }}-openshift-ai
+              namespace: openshift-marketplace
+            status:
+              connectionState:
+                lastObservedState: READY
+        # the ingress-operator creates a servicemeshoperator3 Subscription
+        # with source: redhat-operators, which doesn't exist in disconnected mode.
```
carbonin, line 48: Is this only something that happens when we otherwise install openshift AI?
I guess that's why it's in the openshift-ai plugin?

I worry that if something else happens to go looking for an operator in redhat-operators pointing that catalog source to something openshift-ai specific will cause confusion.

Should we consider mirroring all of redhat-operators now separately from some individual plugin or is that too heavy handed for this fix?

plugins/openshift-ai/plugin.yaml
```diff
@@ -36,7 +36,7 @@ operators:
   - name: servicemeshoperator3
     version: 3.2.2
     channel: stable
-    init_version: 3.2.2
```
eliorerz, line 39: The `init_version` changed from `3.2.2` to `3.1.0` while the `version` stays at `3.2.2`. This will cause a larger mirror range - oc-mirror now will mirror versions from 3.1.0 through 3.2.2 instead of just 3.2.2. 
This is intentional?

plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
```diff
@@ -0,0 +1,73 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  annotations:
+    policy.open-cluster-management.io/categories: CM Configuration Management
+    policy.open-cluster-management.io/controls: CM-2 Baseline Configuration
+    policy.open-cluster-management.io/standards: NIST SP 800-53
+  name: model-catalogsources
```
eliorerz, line 8: The nvidia-gpu uses `nvidia-gpu-catalogsource` for its policy name, which is nice and clear. Here we're using `model-catalogsources` do you think that it might be worth aligning on `openshift-ai-catalogsources` to make it obvious which plugin owns it?

plugins/openshift-ai/plugin.yaml
maorfr: yes, just to be on the safe side

plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
maorfr: yes, absolutely. happy to do it in a follow up PR so we don't rename existing files in this PR. works?

plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
maorfr: significant added mirroring at this point is problematic iiuc. there is more to be done in this area for sure, we need to pick the right timeframe.

plugins/openshift-ai/files/09-policy-catalogsource.yaml.j2
agonzalezrh: short_version_dot

maorfr: https://github.com/rh-ecosystem-edge/enclave/pull/261/commits/b4943925c1389143318da5849ec108945707e22d

agonzalezrh (approved):
lgtm! @maorfr amazing job

maorfr:
amazing job yourself

PR #262, opened 2026-04-16
add comment to pull request with created tarball for testing purposes

maorfr:
we create tarballs as part of pull requests. might as well notify about it.

## Summary by CodeRabbit

* **Chores**
  * Enhanced pull request workflow to automatically include tarball image reference details in PR comments for improved visibility and accessibility during development.


eurijon (commented):
lgtm, don't need to make a link to the tarball, right?

rporres (approved):
LGTM. I wonder if in the long term we're going to get tired of the noise this may introduce, but it's fine for the time being .

PR #287, opened 2026-04-21
Promote trust_custom_ca to standard Ansible post-install flow

rporres:
Move the Proxy/cluster CA trust operation from a CI-only bash script into a proper Ansible task (`playbooks/tasks/trust_custom_ca.yaml`) that runs as part of Phase 4 post-install. The task reads the existing Proxy/cluster trustedCA, guards against re-adding an already-trusted CA (covering disconnected installs where additionalTrustBundle may have set it up already), and merges any prior CA content before patching.

The standalone CI step (`trust-custom-ca`) and its backing script are fully removed from Makefile.ci, the workflow, and the scripts directory.

This solves an undocumented side-effect (self-signed certs work in disconnected mode) and make connected and disconnected work the same way.

## Summary by CodeRabbit

* **Refactor**
  * Moved custom Root CA handling from CI shell scripts into the deployment automation for clearer, centralized management.
* **New Features**
  * Deployment now conditionally adds a provided Root CA to cluster trust when an SSL CA certificate is supplied.
* **Chores**
  * CI no longer runs the prior custom CA trust step or exposes the removed make target.


playbooks/tasks/trust_custom_ca.yaml
```diff
@@ -0,0 +1,84 @@
...
+    kind: Proxy
+    name: cluster
+  register: r_current_proxy
+  retries: "{{ k8s_retries }}"
+  delay: "{{ k8s_delay }}"
+  until:
+    - r_current_proxy is success
+    - r_current_proxy.resources | length > 0
+
+- name: Read existing trusted CA ConfigMap when configured
```
eurijon, line 16: nit: this task is missing the ansible retries and delay params, we might want to add them for consistency across playbooks and robustness

rporres: indeed! fixed in ce149dc

eurijon (approved):

PR #291, opened 2026-04-24
Add OSAC plugin skeleton and E2E CI workflow

eurijon:
Introduce the OSAC operator as an enclave plugin with a minimal deploy task that pulls the operator manifests from a container image and applies them to the cluster. Includes a post-validation check that waits for the controller-manager deployment to become available.

The E2E workflow deploys a connected cluster and runs the plugin on top, triggered by changes to `plugins/osac/` or manual dispatch.

MGMT-24065


maorfr:
consider using `plugin.registries` and the mirrored image that would create

PR #297, opened 2026-04-26
download control binaries to tarball

maorfr:
avoid download dependency at runtime (but adds to the tarball size).

this removes a dependency on download sources which may be unreliable and cause a blocker during install time.
for example, we are downloading clairctl from github, which can not be trusted (you heard me github).


rporres:
Why are we doing this?

maorfr:
> Why are we doing this?

updated the description

rporres (approved):

maorfr:
closing for #464 (#477)

PR #305, opened 2026-04-28
MGMT-23874: fix plugin IDMS/ITMS overwrite during mirror deploy

danielerez:
deploy_plugin.yaml previously overwrote internal IDMS/ITMS during each plugin deployment.
Because oc-mirror generates IDMS/ITMS with the same resource name for every plugin, 
the latest plugin run replaced mirror mappings from earlier runs.
This change keeps plugin mirror entries merged so mappings accumulate across plugins.

I.e. This change isolates plugin mirror resources under a dedicated "-plugins" suffix
and introduces a merge helper that combines mirror entries by source before apply.
The result is an additive behavior: new plugin mirrors are appended without clobbering
existing IDMS/ITMS mappings, while non-plugin flows keep using the original
internal resource path unchanged.

Example resulting resources:
$ oc get idms
NAME
idms-operator-0
idms-operator-0-internal
idms-operator-0-plugins

## Summary by CodeRabbit

* **New Features**
  * CLI to merge oc-mirror-generated mirror entries into cluster mirror resources, preserving existing sources and appending missing entries.
  * Optionally merge plugin-provided mirror sets into consolidated manifests used during deployment.
  * Configurable destination naming/suffixing to avoid mirror name conflicts and select merged vs non-merged manifests.

* **Chores**
  * Added deployment tasks to conditionally run merges and apply merged manifests with retry handling.

### DIFF
--- a/helpers/mirror/merge_mirror_sets.py
+++ b/helpers/mirror/merge_mirror_sets.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python3
+"""
+Merge oc-mirror generated IDMS/ITMS manifests with existing cluster resources.
+
+The merge strategy is keyed by `source` and preserves existing mirror order:
+- Existing mirrors remain first.
+- New mirrors from the current manifest are appended if missing.
+"""
+
+import argparse
+import json
+import subprocess
+import sys
+import time
+
+DEFAULT_SUBPROCESS_TIMEOUT = 30
+
+
+def _format_cmd(argv):
+    return " ".join(argv)
+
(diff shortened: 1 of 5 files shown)


maorfr:
can we consider an approach that every plugin uses a custom catalog? that would remove the concern of overwriting completely. made the same suggestion in https://github.com/rh-ecosystem-edge/enclave/pull/288#issuecomment-4305001170

as for naming, i've created #306 which may help, if we adopt the above suggestion.

PR #308, opened 2026-04-28
only mirror plugin when operators is not an empty list

maorfr:
otherwise what would be mirrored?

## Bug Fixes

* Enhanced plugin mirroring logic to execute when plugins include operator or additional images, improving configuration flexibility
* Strengthened operator installation validation by requiring list elements rather than just variable presence
* Improved catalog configuration handling to support additional images scenarios

### DIFF
--- a/playbooks/tasks/deploy_plugin.yaml
+++ b/playbooks/tasks/deploy_plugin.yaml
@@ -119,6 +119,7 @@
     file: mirror_plugin.yaml
   when:
     - plugin.mirror | default('none') == 'plugin'
+    - plugin.operators | default([]) | length > 0
     - disconnected | default(true) | bool
   tags: mirror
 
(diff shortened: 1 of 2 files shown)


javipolo (approved):

eliorerz (changes requested):
A plugin could theoretically have `mirror: plugin` with no operators but with `additionalImages` that need to be mirrored to the registry. After this PR, such a plugin would skip mirroring entirely because `plugin.operators | length > 0` is false.

maorfr:
> A plugin could theoretically have `mirror: plugin` with no operators but with `additionalImages` that need to be mirrored to the registry. After this PR, such a plugin would skip mirroring entirely because `plugin.operators | length > 0` is false.

great catch @eliorerz! updated the logic in https://github.com/rh-ecosystem-edge/enclave/pull/308/commits/7b31f9e7abad5541d3b0a3ed98d041dd48c0605a

PR #309, opened 2026-04-28
add default operator source to configure operator and rename default catalog

maorfr:

we use `cs-redhat-operator-index-v4-20` which is the default catalog name when not defining a target catalog in ImageSetConfiguration. this PR uses the variable `mirror_rh_operator_catalog` as the name and target catalog for the core mirroring.

we're introducing a default operator source that is determined by the execution flow with a `default_operator_source` variable, and we keep `operator.source` as a way to override default values.

for plugins we set the default to be the per-plugin catalog and keeping overrides while we establish a unified approach for mirroring plugins.

## Summary by CodeRabbit

* **Documentation**
  - Simplified operator configuration examples in the deployment guide and plugin architecture docs.

* **Configuration**
  - Switched default operator catalog source to use the mirrored operator index and added support for directing mirrored operators to a target catalog path.
  - Improved operator source selection for disconnected installs to allow configurable defaults.

### DIFF
--- a/files/catalogsource-configuration/10-policy.yaml.j2
+++ b/files/catalogsource-configuration/10-policy.yaml.j2
@@ -41,5 +41,5 @@ spec:
               name: {{ mirror_rh_operator_catalog }}
               namespace: openshift-marketplace
             spec:
-              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/redhat-operator-index:v{{ short_version_dot }}
+              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/{{ mirror_rh_operator_catalog }}:v{{ short_version_dot }}
               sourceType: grpc
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -91,7 +91,7 @@
         installPlanApproval: Automatic
         channel: "{{ operator.channel }}"
         name: "{{ operator.name }}"
-        source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
+        source: "{{ operator.source | default(default_operator_source) if (disconnected | default(true)) else 'redhat-operators' }}"
         sourceNamespace: openshift-marketplace
   register: r_sub_exists
   retries: "{{ k8s_retries }}"
@@ -114,7 +114,7 @@
         installPlanApproval: Manual
         channel: "{{ operator.channel }}"
         name: "{{ operator.name }}"
-        source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
+        source: "{{ operator.source | default(default_operator_source) if (disconnected | default(true)) else 'redhat-operators' }}"
         sourceNamespace: openshift-marketplace
         startingCSV: "{{ operator.csvNames | default([]) | first | default(operator.name) }}.v{{ operator.version | replace('+', '-') }}"
   register: r_sub_exists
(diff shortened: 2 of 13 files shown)


playbooks/tasks/configure_operator.yaml
```diff
@@ -91,7 +91,7 @@
         installPlanApproval: Automatic
         channel: "{{ operator.channel }}"
         name: "{{ operator.name }}"
-        source: "{{ operator.source if (disconnected | default(true)) else 'redhat-operators' }}"
+        source: "{{ mirror_rh_operator_catalog if (disconnected | default(true)) else 'redhat-operators' }}"
```
eliorerz: This hardcodes all operators to `mirror_rh_operator_catalog`, but `mirror: plugin` operators have per-plugin `CatalogSources` with different names (e.g., cs-mirror-redhat-operators-odf-...). 
Should this be something like `operator.source | default(mirror_rh_operator_catalog)` so plugins can still override the source? Otherwise standalone plugin deployment in disconnected mode will point `Subscriptions` at a `CatalogSource` that doesn't have the packages.

eliorerz (commented):
Looks great, Few things to think about:
 - Maybe it worth merging some of the commit messages (e.g. the `keep fixes` or `remove quotes` commits)
 - Regarding the `source` removal, maybe it worth keeping it as an optional field with a note that it defaults to `mirror_rh_operator_catalog`?

playbooks/tasks/configure_operator.yaml
maorfr: introduced a `default_operator_source` variable that is set when calling configure_operator. also kept `operator.source` for overrides. great ideas!

files/catalogsource-configuration/10-policy.yaml.j2
```diff
@@ -41,5 +41,5 @@ spec:
               name: {{ mirror_rh_operator_catalog }}
               namespace: openshift-marketplace
             spec:
-              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/redhat-operator-index:v{{ short_version_dot }}
+              image: registry-quay-quay-enterprise.apps.{{ clusterName }}.{{ baseDomain }}/redhat/{{ mirror_rh_operator_catalog }}:v{{ short_version_dot }}
```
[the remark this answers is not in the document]
maorfr: good thought, will keep it for a future PR.

playbooks/tasks/configure_operators.yaml
```diff
@@ -3,6 +3,8 @@
   block:
 
     - name: Include configure_operator
+      vars:
+        default_operator_source: "cs-{{ mirror_rh_operator_catalog }}-v4-20"
```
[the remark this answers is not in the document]
maorfr: good thought, will do it in a later PR.

rporres:
Still wrapping my head around this after a good while reading it 😂 

I think you're going to tackle the 4-20 hardcoding in a subsequent PR, to derive it from `rh_operator_catalog_version`

All in all, LGTM, but I'll defer to @eliorerz for the final confirmation

eliorerz (approved):

playbooks/tasks/deploy_plugin.yaml
```diff
@@ -139,6 +155,8 @@
   tags: mirror
 
 - name: Install plugin operators
+  vars:
+    default_operator_source: "{{ catalog_mirror }}"
```
maorfr, line 159: bug fixed in https://github.com/rh-ecosystem-edge/enclave/pull/318/changes#diff-6f9767dfd8e0bfd6da8ec3238f1c1b382ecd2863fd563842cb38631ad966250e

PR #310, opened 2026-04-28
broaden default Quay OAuth scopes

helsus:
## Description

Adds `user:admin` and `org:admin` to the default scopes. `oauth_setup.yaml` skips org/app creation when `quay-oauth-credentials` exists and regenerates the token only when scopes drift. `quay-oauth-sync` tag in `06-day2.yaml` covers existing clusters.

## Open question

- [x] Is `06-day2.yaml` the right place to add sync for upgrading existing clusters or should it live in a standalone playbook?

## Testing:

- [ ] Fresh install -- new scopes
- [ ] Existing cluster with old scopes -- sync regenerates
- [ ] Idempotency -- re-running is a no-op
- [ ] Order-insensitive scope match -- no regen
- [ ] Legacy Secret missing the scopes key -- sync handles it
- [ ] Token actually carries the new scopes/permissions

## Summary by CodeRabbit

* **New Features**
  * Expanded Quay OAuth application scopes to include user administrator and organization administrator permissions.

* **Documentation**
  * Introduced new runbook documenting the process for updating Quay OAuth access tokens with verification and validation steps.

### DIFF
--- a/operators/quay-operator/oauth_setup.yaml
+++ b/operators/quay-operator/oauth_setup.yaml
@@ -9,7 +9,28 @@
 # This file sets the following variables:
 #   - quay_oauth_client_id
 #   - quay_oauth_client_secret
-#   - quay_oauth_token
+#   - quay_oauth_token (only when a token is (re)generated)
+
+- name: Read existing OAuth credentials Secret
+  no_log: true
+  kubernetes.core.k8s_info:
+    api_version: v1
+    kind: Secret
+    name: quay-oauth-credentials
+    namespace: quay-enterprise
+  register: r_quay_oauth_secret_existing
+
+- name: Determine whether OAuth application is already provisioned
+  ansible.builtin.set_fact:
+    quay_oauth_secret_exists: "{{ r_quay_oauth_secret_existing.resources | default([]) | length > 0 }}"
+
+- name: Reuse credentials from existing Secret
+  no_log: true
+  when: quay_oauth_secret_exists
+  ansible.builtin.set_fact:
+    quay_oauth_client_id: "{{ r_quay_oauth_secret_existing.resources[0].data['client-id'] | b64decode }}"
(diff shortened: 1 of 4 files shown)


pgodowski:
@oourfali @maorfr can you please take a look? 
thanks

operators/quay-operator/oauth_setup.yaml
```diff
@@ -160,6 +185,7 @@
   when:
     - quay_oauth_client_id is defined
     - quay_oauth_client_id | length > 0
+    - not quay_oauth_secret_exists or (quay_oauth_current_scopes.split() | sort) != (quayOAuthApp.scopes.split() | sort)
```
maorfr: what happens to current token if scopes have changed?

maorfr (changes requested):
this PR introduces a configuration change together with somewhat of a migration to put it in place.

not sure this is a desired pattern at this point.

i suggest to submit the configuration change only (together with a document perhaps). generally speaking, i think `quay-oauth` would be a nice plugin (as in, extract this entire functionality to a plugin).

chunlongliang-ibm:
is the oauth token only available to MSP admin users, and  NOT accessible to tenant?

helsus:
> is the oauth token only available to MSP admin users, and NOT accessible to tenant?

Yes, it is only available to MSP admins.

operators/quay-operator/oauth_setup.yaml
helsus: It will remain in Quay. I haven't found a reliable way to revoke it via automation, so I'll document how to revoke it manually via UI.

helsus:
> this PR introduces a configuration change together with somewhat of a migration to put it in place.
> 
> not sure this is a desired pattern at this point.
> 
> i suggest to submit the configuration change only (together with a document perhaps). generally speaking, i think `quay-oauth` would be a nice plugin (as in, extract this entire functionality to a plugin).

Good idea. I've updated the PR to include only the config and doc on how to reissue the token on live clusters.

maorfr (approved):

PR #306, opened 2026-04-28
add name metadata to plugin image set configuration

maorfr:
this is intended to be used to distinct plugin mirroring

## Summary by CodeRabbit

* **Chores**
  * Enhanced resource configuration by introducing proper identification metadata for improved resource management and organization.

PR #312, opened 2026-04-28
MGMT-23874: fix plugin IDMS/ITMS overwrite during mirror deploy

danielerez:
Each plugin mirror run used oc-mirror output whose ImageDigestMirrorSet and ImageTagMirrorSet objects reused generic names, so applying Quay disconnected mirrors could overwrite or collide with other plugins or the core flow.

Landing-zone manifests are copied and renamed to stable per-plugin names (idms-plugin-[name], itms-plugin-[name]).
Quay Enterprise follow-on mirrors use matching internal names (idms-plugin-[name]-internal, itms-plugin-[name]-internal).

Example of resulting resources:
$ oc get idms
NAME
idms-operator-0-plugin-lso
idms-operator-0-plugin-lso-internal
idms-operator-0-plugin-mtv
idms-operator-0-plugin-mtv-internal

## Summary by CodeRabbit

* **New Features**
  * Plugin mirror support now generates plugin-specific mirror manifests and unique resource names so each plugin’s mirrors are isolated and applied safely.

* **Chores**
  * Refactored mirror manifest workflow to use mode-dependent naming and dynamic manifest selection, improving flexibility when deploying mirror resources in plugin mode.

### DIFF
--- a/playbooks/tasks/mirror_plugin.yaml
+++ b/playbooks/tasks/mirror_plugin.yaml
@@ -85,12 +85,71 @@
           Full log: {{ _oc_mirror_plugin_log }}
 
 - name: Apply updated mirror manifests to cluster
-  ansible.builtin.shell: |
-    {{ workingDir }}/bin/oc apply \
-      -f {{ workingDir }}/config/oc-mirror-workspace/working-dir/cluster-resources/ \
-      --server-side --force-conflicts
-  environment:
-    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+  block:
+    - name: Set landing-zone cluster resources directory
+      ansible.builtin.set_fact:
+        lz_cluster_resources_dir: "{{ workingDir }}/config/oc-mirror-workspace/working-dir/cluster-resources"
+
+    - name: Stat landing-zone IDMS source
+      ansible.builtin.stat:
+        path: "{{ lz_cluster_resources_dir }}/idms-oc-mirror.yaml"
+      register: lz_idms_src_stat
+
+    - name: Stat landing-zone ITMS source
+      ansible.builtin.stat:
+        path: "{{ lz_cluster_resources_dir }}/itms-oc-mirror.yaml"
+      register: lz_itms_src_stat
+
+    # Dedicated IDMS/ITMS per plugin; oc-mirror overwrites generic idms-oc-mirror.yaml — remove after copy.
+    # oc-mirror v2 may emit multiple resources per file; names must stay unique (idms-<slice>-plugin-<name>).
+    # Quay apply uses idms-<slice>-plugin-<name>-internal (quay_disconnected_mirrors.yaml).
+    - name: Copy LZ IDMS to plugin-specific manifest
+      ansible.builtin.copy:
+        src: "{{ lz_cluster_resources_dir }}/idms-oc-mirror.yaml"
+        dest: "{{ lz_cluster_resources_dir }}/idms-oc-mirror-plugin-{{ plugin_name }}.yaml"
+        remote_src: true
+      when: lz_idms_src_stat.stat.exists
+
+    - name: Rename ImageDigestMirrorSet names in plugin LZ IDMS manifest
+      ansible.builtin.replace:
+        path: "{{ lz_cluster_resources_dir }}/idms-oc-mirror-plugin-{{ plugin_name }}.yaml"
+        regexp: '^(\s*name:\s*)idms-([^\s]+)\s*$'
+        replace: '\1idms-\2-plugin-{{ plugin_name }}'
+      when: lz_idms_src_stat.stat.exists
+
+    - name: Remove generic LZ IDMS after plugin-specific copy
+      ansible.builtin.file:
+        path: "{{ lz_cluster_resources_dir }}/idms-oc-mirror.yaml"
+        state: absent
+      when: lz_idms_src_stat.stat.exists
+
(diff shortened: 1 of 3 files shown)


maorfr:
can we consider an approach that every plugin uses a custom catalog? that would remove the concern of overwriting completely. made the same suggestion in https://github.com/rh-ecosystem-edge/enclave/pull/288#issuecomment-4305001170

as for naming, i've created https://github.com/rh-ecosystem-edge/enclave/pull/306 which may help, if we adopt the above suggestion.

https://github.com/rh-ecosystem-edge/enclave/pull/305#issuecomment-4334130279

danielerez:
> can we consider an approach that every plugin uses a custom catalog? that would remove the concern of overwriting completely. made the same suggestion in [#288 (comment)](https://github.com/rh-ecosystem-edge/enclave/pull/288#issuecomment-4305001170)
> 
> as for naming, i've created #306 which may help, if we adopt the above suggestion.
> 
> [#305 (comment)](https://github.com/rh-ecosystem-edge/enclave/pull/305#issuecomment-4334130279)

Yeah, I've actually considered exactly that:)
But assumed it might be cleaner/clearer to handle a single IDMS/ITMS resource for all plugins.
Any way, it should simplify the logic and indeed reduce risks. So I'll look into it.

playbooks/tasks/mirror_plugin.yaml
```diff
@@ -87,12 +87,47 @@
...
+      ansible.builtin.include_tasks:
+        file: merge_mirror_manifests.yaml
+      vars:
+        merge_mirror_enabled: "{{ quay_merge_plugin_mirrors | default(false) | bool }}"
+        merge_mirror_items:
+          - api_kind: ImageDigestMirrorSet
+            oc_resource: imagedigestmirrorset
+            spec_key: imageDigestMirrors
+            input_manifest: "{{ lz_cluster_resources_dir }}/idms-oc-mirror.yaml"
+            output_manifest: "{{ lz_cluster_resources_dir }}/idms-oc-mirror.yaml"
```
eliorerz: Looks like that the "LZ merge" uses the same path for input and output (`idms-oc-mirror.yaml`). The Quay Enterprise flow uses a separate `*-merged.json` path. Any reason not to do the same for LZ? It would avoid the YAML-in/JSON-out surprise and make it clearer what's the original vs. merged artifact.

playbooks/tasks/merge_mirror_manifests.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - --oc-resource
+      - "{{ item.oc_resource }}"
+      - --spec-key
+      - "{{ item.spec_key }}"
+      - --apply
+      - --max-attempts
+      - "{{ merge_mirror_max_attempts | default(5) }}"
+      - --retry-delay
+      - "{{ merge_mirror_retry_delay | default(1.0) }}"
+  changed_when: false
```
eliorerz: I think that with `--apply` the script modifies the cluster resources. Should this be `changed_when: true` (or check script output) so Ansible's change tracking is accurate?

maorfr:
> > can we consider an approach that every plugin uses a custom catalog? that would remove the concern of overwriting completely. made the same suggestion in [#288 (comment)](https://github.com/rh-ecosystem-edge/enclave/pull/288#issuecomment-4305001170)
> > as for naming, i've created #306 which may help, if we adopt the above suggestion.
> > [#305 (comment)](https://github.com/rh-ecosystem-edge/enclave/pull/305#issuecomment-4334130279)
> 
> Yeah, I've actually considered exactly that:) But assumed it might be cleaner/clearer to handle a single IDMS/ITMS resource for all plugins. Any way, it should simplify the logic and indeed reduce risks. So I'll look into it.

started illustrating what that might look like in #316.

danielerez:
> #316

Thanks, as suggested, changed this PR to generate a per-plugin IDMS/ITMS.

playbooks/tasks/mirror_plugin.yaml
danielerez: Thanks, changed the logic to apply a per-plugin resource as suggested, so not relevant any more:)

playbooks/tasks/merge_mirror_manifests.yaml
danielerez: Same, removed the merge logic already.

maorfr:
> > #316
> 
> Thanks, as suggested, changed this PR to generate a per-plugin IDMS/ITMS.

hope you don't mind, i also tried something: #318. trying to be a good oc-mirror user. do you think it makes sense?

danielerez:
> > > #316
> > 
> > 
> > Thanks, as suggested, changed this PR to generate a per-plugin IDMS/ITMS.
> 
> hope you don't mind, i also tried something: #318. trying to be a good oc-mirror user. do you think it makes sense?

Sure, added a comment there. But it looks complimentary to this change:) Still need this mirror set handling.

maorfr:
> Sure, added a comment there. But it looks complimentary to this change:) Still need this mirror set handling.

i agree. i'm assuming that this PR may slightly change in implementation as a result of #318?

danielerez:
> > Sure, added a comment there. But it looks complimentary to this change:) Still need this mirror set handling.
> 
> i agree. i'm assuming that this PR may slightly change in implementation as a result of #318?

Actually not really, there's no special handling for core plugins in this change. Tested it locally with #318 and seems to work fine.

maorfr (approved):
LGTM

i have much more to learn

PR #298, opened 2026-04-26
mirror plugin make target and extract section

maorfr:
part of https://redhat.atlassian.net/browse/OSAC-875

add target to be used as part of a plugin upgrade process.

generally, an upgrade process is mirroring + upgrade. this PR adds a capability to mirror a plugin without upgrading any aspects of it.

## Summary by CodeRabbit

* **New Features**
  * Added a new make target to run plugin mirroring from the command line (requires specifying which plugin).
  * Added a dedicated mirroring workflow that supports disconnected environments and mirrors plugin operator images to a private registry.

* **Refactor**
  * Simplified mirroring flow for improved reliability: retries, progress logging, readiness waits, explicit cleanup of temporary registry config, and clearer error reporting.

PR #318, opened 2026-04-29
core plugin mirror to custom catalog

maorfr:

follow up on #261 and #309

the goal of the PR is to mirror core plugins to a custom per-plugin catalog, and thus making this method consistent for all plugins. this PR may also serve as a stepping stone towards enabling day2 installations of core plugins.

the problem is that core plugins are currently mirrored into the default catalog. in day2 when we install only the plugin, we don't have the full context of the default catalog and it will be overridden.

in this PR we are changing the imagesetconfiguration template to create a catalog per core plugin. this enables us to install a core plugin on day0 with a custom catalog, which will be consistent with a day2 installation.

to change the imagesetconfiguration template we are updating the `_core_plugin_operators` from a list to a dict (with the plugin name as the key), allowing us to iterate over plugins and create custom catalogs.

we update the list of custom manifests passed to openshift-install to find all custom catalog sources.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized operator catalog output to emit per-plugin operator catalogs rather than a single combined list.
* **Chores**
  * Switched to runtime discovery of mirror manifest files instead of a fixed file list.
  * Updated operator catalog source naming to use plugin-specific mirrored catalogs and versioned source strings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/318)

### DIFF
--- a/playbooks/tasks/mirror_registry.yaml
+++ b/playbooks/tasks/mirror_registry.yaml
@@ -36,10 +36,46 @@
     content: "{{ pullSecretInternal }}"
     dest: "{{ workingDir }}/config/pull-secret.internal.json"
 
+# We iterate over catalog_operators keys, which are the list of plugins + "core".
+# For each item, we copy the Red Hat Index image to an intermediate image in the mirror registry,
+# to be used as the source catalog image in the ImageSetConfiguration of oc-mirror.
+# This is required to make oc-mirror output a Catalog Source file for each item,
+# which we later use during cluster provisioning.
+# This is done to separate mirroring of plugins from the core operators,
+# and to support plugin operations in day 2.
+# The destination mirror registry does not support signatures and TLS,
+# so we work around by using remove-signatures, insecure-policy and dest-tls-verify.
+- name: Copy Red Hat Index Image for core and plugin operators
+  ansible.builtin.shell: |
+    skopeo copy \
+      --all \
+      --remove-signatures \
+      --insecure-policy \
+      --authfile {{ pullSecretPath }} \
+      --dest-tls-verify=false \
+      --retry-times {{ ocMirrorRetryTimes }} \
+      --retry-delay {{ ocMirrorRetryDelay }} \
+      docker://{{ rh_operator_catalog }}:{{ rh_operator_catalog_version }} \
+      docker://{{ quayHostname }}:8443/redhat/redhat-operator-index{{ '-' ~ item }}:{{ rh_operator_catalog_version }}
+  loop: "{{ catalog_operators.keys() | list }}"
+  loop_control:
+    label: "{{ item }}"
+  register: skopeo_result
+  changed_when: "'Copying blob' in skopeo_result.stdout"
+  retries: "{{ 1 if (mirror_dry_run | default(false) | bool) else ocMirrorAnsibleRetries }}"
+  delay: "{{ 0 if (mirror_dry_run | default(false) | bool) else ocMirrorAnsibleDelay }}"
+  until: skopeo_result.rc == 0
+
 - name: Copy imagesetconfiguration.yaml
   ansible.builtin.template:
     src: ../../templates/imagesetconfiguration.yaml.j2
     dest: "{{ workingDir }}/config/imagesetconfiguration.yaml"
+
+- name: Replace registry.redhat.io with redhat-operator-index in mirror-registry
+  ansible.builtin.replace:
+    path: "{{ workingDir }}/config/imagesetconfiguration.yaml"
+    regexp: 'catalog: {{ rh_operator_catalog }}'
+    replace: "catalog: {{ quayHostname }}:8443/redhat/redhat-operator-index"
 
 - name: Ensure oc-mirror log directory exists
   ansible.builtin.file:
@@ -62,6 +98,8 @@
           {{ '--dry-run' if (mirror_dry_run | default(false) | bool) else '' }} \
           --log-level {{ ocMirrorLogLevel }} \
           --authfile {{ pullSecretPath }} \
+          --ignore-release-signature \
+          --src-tls-verify=false \
           --dest-tls-verify=false \
           -c {{ workingDir }}/config/imagesetconfiguration.yaml \
           --workspace file://{{ workingDir }}/config/oc-mirror-workspace \
--- a/playbooks/tasks/mirror_registry.yaml
+++ b/playbooks/tasks/mirror_registry.yaml
@@ -56,7 +56,7 @@
       --retry-times {{ ocMirrorRetryTimes }} \
       --retry-delay {{ ocMirrorRetryDelay }} \
       docker://{{ rh_operator_catalog }}:{{ rh_operator_catalog_version }} \
-      docker://{{ quayHostname }}:8443/redhat/redhat-operator-index{{ '-' ~ item }}:{{ rh_operator_catalog_version }}
+      docker://{{ quayHostname }}:8443/redhat/redhat-operator-index{{ (item == 'core') | ternary('', '-' ~ item) }}:{{ rh_operator_catalog_version }}
   loop: "{{ catalog_operators.keys() | list }}"
   loop_control:
     label: "{{ item }}"
--- a/docs/PLUGIN_ARCHITECTURE.md
+++ b/docs/PLUGIN_ARCHITECTURE.md
@@ -209,6 +209,8 @@ Plugins with `mirror: core` have their operators collected by `collect_core_plug
 
 Operators with `csvMirror: true` have their `csvNames` entries added as separate packages in the image set.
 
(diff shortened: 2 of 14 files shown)


maorfr:
@coderabbitai review

eliorerz:
@maorfr Did you tested this installation scenario? If not please test it before merging this PR. We should test disconnected LVMS installation and on day2 install ODF. The CI does not test this so it might better to move it to WIP until it is fully tested.
I also created https://github.com/rh-ecosystem-edge/enclave/pull/288 days ago but the day2 installation broke the installation.

maorfr:
> @maorfr Did you tested this installation scenario? If not please test it before merging this PR. We should test disconnected LVMS installation and on day2 install ODF. The CI does not test this so it might better to move it to WIP until it is fully tested. I also created #288 days ago but the day2 installation broke the installation.

will do. do you have any input on the approach?

danielerez:
> > @maorfr Did you tested this installation scenario? If not please test it before merging this PR. We should test disconnected LVMS installation and on day2 install ODF. The CI does not test this so it might better to move it to WIP until it is fully tested. I also created #288 days ago but the day2 installation broke the installation.
> 
> will do. do you have any input on the approach?

So what is actually the difference now between 'core' and 'plugin'?
I mean, maybe we don't need the 'core' terminology? It could also simplify the logic a bit:)

maorfr:
> So what is actually the difference now between 'core' and 'plugin'? I mean, maybe we don't need the 'core' terminology? It could also simplify the logic a bit:)

that is exactly right! the next PR after this would be to simlify this logic. i've added a safety net towards this change in #298.

ref: https://github.com/rh-ecosystem-edge/enclave/pull/288#issuecomment-4305001170

eliorerz:
> will do. do you have any input on the approach?

Sure, You can follow the disconnected LVMS installation flow or do it manually. Once the cluster is installed you should deploy the ODF plugin (using make or Ansible). If you need a valid storage for ODF you can use the ceph installation target (see `setup-ceph` Makefile target) and configure it using `export ODF_EXTERNAL_CONFIG='<odf_external_configuration>'` which will be exist in the LZ at: `/home/cloud-user/ceph-config/odf_external_config.json`

danielerez:
e2e job seems to fail on oc-mirror, maybe needs a rebase?
```
TASK [oc-mirror failed] ********************************************************
Wednesday 06 May 2026  16:51:28 +0000 (0:00:00.322)       0:02:44.841 ********* 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "oc-mirror for plugin lvms (Quay Enterprise) failed after 10 attempts.\n\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : 👋 Hello, welcome to oc-mirror\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : ⚙️  setting up the environment for you...\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : ⚙️  environment version: 4.20.0-202602091941.p2.ga8c29a3.assembly.stream.el9-a8c29a3\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : 🔀 workflow mode: mirrorToMirror \n2026/05/06 16:51:28 \u001b[1;91m [ERROR] \u001b[0m : [Executor] checking registry \"registry-quay-quay-enterprise.apps.ecd-801ee448.ecd-801ee448.lab\" access: failed to authenticate: pinging container registry registry-quay-quay-enterprise.apps.ecd-801ee448.ecd-801ee448.lab: received unexpected HTTP status: 503 Service Unavailable \n\nFull log: /home/cloud-user/logs/oc-mirror-plugin-lvms-quay.progress.1778086182.log"}
```

maorfr:
ah! it's trying to mirror to quay before quay is deployed.

danielerez:
Same oc-mirror failure:
```
TASK [oc-mirror failed] ********************************************************
Wednesday 06 May 2026  16:51:28 +0000 (0:00:00.322)       0:02:44.841 ********* 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "oc-mirror for plugin lvms (Quay Enterprise) failed after 10 attempts.\n\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : 👋 Hello, welcome to oc-mirror\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : ⚙️  setting up the environment for you...\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : ⚙️  environment version: 4.20.0-202602091941.p2.ga8c29a3.assembly.stream.el9-a8c29a3\n2026/05/06 16:51:28 \u001b[1;94m [INFO] \u001b[0m  : 🔀 workflow mode: mirrorToMirror \n2026/05/06 16:51:28 \u001b[1;91m [ERROR] \u001b[0m : [Executor] checking registry \"registry-quay-quay-enterprise.apps.ecd-801ee448.ecd-801ee448.lab\" access: failed to authenticate: pinging container registry registry-quay-quay-enterprise.apps.ecd-801ee448.ecd-801ee448.lab: received unexpected HTTP status: 503 Service Unavailable \n\nFull log: /home/cloud-user/logs/oc-mirror-plugin-lvms-quay.progress.1778086182.log"}
```

I think it fails as the LVMS is mandatory for the internal Quay registry.
I.e. we can't mirror that plugin since the Quay registry isn't running yet.
Maybe we should just deploy the lvms plugin before deploying Quay?
Or just have a specific flag? Which essentially can be something like 'core':)

Update: oops, missed the last comment, yeah basically this: https://github.com/rh-ecosystem-edge/enclave/pull/318#issuecomment-4396338664

maorfr:
will save the change of flow to a later PR and focused this on the custom catalog implementation.

playbooks/tasks/mirror_registry.yaml
```diff
@@ -62,6 +89,8 @@
           {{ '--dry-run' if (mirror_dry_run | default(false) | bool) else '' }} \
           --log-level {{ ocMirrorLogLevel }} \
           --authfile {{ pullSecretPath }} \
+          --ignore-release-signature \
```
danielerez, line 101: Do we have to skip signature verification? I.e. do we need to support envs that don't have the release keys in trust bundle?

playbooks/tasks/mirror_registry.yaml
```diff
@@ -36,11 +36,38 @@
     content: "{{ pullSecretInternal }}"
     dest: "{{ workingDir }}/config/pull-secret.internal.json"
 
+- name: Copy Red Hat Index Image for default and plugin operators
+  ansible.builtin.shell: |
+    skopeo copy \
+    --all \
+    --remove-signatures \
```
danielerez: Can we keep the signatures? Or is it failing? Maybe missing release keys in bundle?

maorfr: we remove signatures because the destination quay does not support signatures

playbooks/tasks/mirror_registry.yaml
maorfr: we ignore signatures because we remove signatures in the skopeo copy step

playbooks/tasks/mirror_registry.yaml
```diff
@@ -36,11 +36,38 @@
     content: "{{ pullSecretInternal }}"
     dest: "{{ workingDir }}/config/pull-secret.internal.json"
 
+- name: Copy Red Hat Index Image for default and plugin operators
```
rporres: We need either a nice comment or a good chunk of documentation to explain what's going on here and why we need this workaround 😅

playbooks/tasks/mirror_registry.yaml
```diff
@@ -36,11 +36,38 @@
...
+  ansible.builtin.shell: |
+    skopeo copy \
+    --all \
+    --remove-signatures \
+    --insecure-policy \
```
rporres: why do we need these two options? A comment would be helpful.

playbooks/tasks/mirror_registry.yaml
```diff
@@ -36,11 +36,38 @@
...
+    --authfile {{ pullSecretPath }} \
+    --dest-tls-verify=false \
```
rporres: same with this option.

rporres (commented):
This is great stuff. It should fix the core plugin upgrade issues. It needs more documentation to understand the implementation choices and workarounds and something in the description that states the current problems it solves.

rporres (approved):
I'm not 100% sure that signatures won't come back to haunt us, but at least we're honoring them when we pull from quay.io.

For the moment, this is the only way that we have to make enclave truly upgradeable, so let's move on with it.

PR #327, opened 2026-05-04
configure seed operators first

maorfr:
define seed operators as operators that are installed before anything else.

open to discuss the right word, core seems to be widely used.

## Summary by CodeRabbit

* **New Features**
  * Operators can be marked as "seed" to ensure they're deployed before others.

* **Chores**
  * Deployment workflow split into a seed-operators phase and a follow-up phase for remaining operators.
  * Operator schema updated to include the new seed property.

* **Tests**
  * Verification checks updated to expect separate tasks for seed and remaining operator configuration.


rporres:
How does this work with the idea of moving all operators to plugins?

rporres (approved):
I understand this is a way to establish an order in operator installation. The non-seed operators will be gradually be moved into foundation plugins.

PR #328, opened 2026-05-05
extract observability plugin

maorfr:
trying to gradually decompose the core bundle to plugins


rporres:
Is it the plan to move all operators to plugin architecture?

PR #333, opened 2026-05-07
cluster upgrade cli subcommand

rporres:
Based on the work in #123

Tests and dependencies scaffolding missing, will do that in a subsequent PR.

## Summary by CodeRabbit

* **New Features**
  * Added a global --log-level CLI option to control command logging.
  * Replaced a placeholder command with a working cluster-version command that requires a version, supports --dry-run/--no-dry-run, and configurable timeout/sleep intervals.
  * Introduced a cluster upgrade workflow: semantic-version validation (prevents downgrades), available-updates checks, operator-readiness verification, wait-for-completion behavior, and dry-run support.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/333)

### DIFF
--- a/reconcile/cluster_upgrade.py
+++ b/reconcile/cluster_upgrade.py
@@ -179,7 +179,7 @@ def wait_for_resource_status(
                     kind,
                     name,
                     "-o",
-                    f"jsonpath='{{.status.{status_field}}}'",
+                    f"jsonpath={{.status.{status_field}}}",
                 ]
             )
         except TimeoutError as e:
--- a/reconcile/cluster_upgrade.py
+++ b/reconcile/cluster_upgrade.py
@@ -367,6 +367,8 @@ def check_cluster_operators_ready() -> tuple[bool, list[str]]:
                 issues.append(f"Cluster Operator {co_name} is not Available")
             if condition_type == "Upgradeable" and condition_status == "False":
                 issues.append(f"Cluster Operator {co_name} is not Upgradeable")
+            if condition_type == "Progressing" and condition_status == "True":
+                issues.append(f"Cluster Operator {co_name} is moving from one steady state to another")
 
     return (len(issues) == 0, issues)
 
--- a/reconcile/cluster_upgrade.py
+++ b/reconcile/cluster_upgrade.py
@@ -1,0 +1,483 @@
+#!/usr/bin/env python3
+
+import json
+import logging
+import math
+import subprocess
+import time
+from typing import Any, Optional
+
+from packaging.version import InvalidVersion, Version
+
+
+class ClusterUpgradeError(Exception):
+    """Base exception for cluster upgrade operations."""
+
+    pass
+
+
+class InvalidVersionError(ClusterUpgradeError):
+    """Raised when a version string does not conform to semantic versioning."""
+
+    def __init__(self, version: str, reason: str):
+        self.version = version
+        self.reason = reason
+        super().__init__(f"Invalid version '{version}': {reason}")
+
+
+class VersionDowngradeError(ClusterUpgradeError):
+    """Raised when attempting to downgrade to an older version."""
+
+    def __init__(self, current: str, desired: str):
+        self.current_version = current
+        self.desired_version = desired
+        super().__init__(
+            f"Cannot downgrade from version {current} to {desired}. "
+            f"Current version is newer than desired version."
+        )
+
+
+class UpdateGraphUnavailableError(ClusterUpgradeError):
+    """Raised when the cluster's update graph is not available."""
+
+    def __init__(self):
+        super().__init__(
+            "Cluster update graph is unavailable (availableUpdates is null). "
+            "The cluster may need to sync with the update service."
+        )
+
+
+class VersionNotAvailableError(ClusterUpgradeError):
+    """Raised when the desired version is not in available updates."""
+
+    def __init__(self, desired: str, available: list[str]):
+        self.desired_version = desired
+        self.available_versions = available
+        super().__init__(
+            f"Version {desired} is not available for upgrade. "
+            f"Available versions: {', '.join(available)}"
+        )
+
+
+class ClusterOperatorsNotReadyError(ClusterUpgradeError):
+    """Raised when cluster operators are not ready for upgrade."""
+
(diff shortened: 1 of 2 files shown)


rporres:
An example debug run in a SNO cluster: 
[output.txt](https://github.com/user-attachments/files/27525132/output.txt)

reconcile/cluster_upgrade.py
```diff
@@ -0,0 +1,483 @@
+#!/usr/bin/env python3
+
+import json
+import logging
+import math
+import subprocess
+import time
+from typing import Any, Optional
+
+from packaging.version import InvalidVersion, Version
```
eliorerz: `packaging` is imported but the PR description says "dependencies scaffolding missing, will do that in a subsequent PR." This will fail at runtime if packaging isn't installed. Not sure that we want to merge code that will not work

reconcile/cluster_upgrade.py
```diff
@@ -0,0 +1,483 @@
...
+    while True:
+        try:
+            result = run_oc_command(
+                [
+                    "oc",
+                    "get",
+                    kind,
+                    name,
+                    "-o",
+                    f"jsonpath='{{.status.{status_field}}}'",
```
eliorerz: I don't think you need the single quotes around the jsonpath value
```suggestion
                    f"jsonpath={{.status.{status_field}}}",
```

reconcile/cluster_upgrade.py
```diff
@@ -0,0 +1,483 @@
...
+        co_status_conditions = co.get("status", {}).get("conditions", [])
+        for condition in co_status_conditions:
+            condition_status = condition.get("status", "")
+            condition_type = condition.get("type", "")
+            if condition_type == "Degraded" and condition_status == "True":
+                issues.append(f"Cluster Operator {co_name} is Degraded")
+            if condition_type == "Available" and condition_status == "False":
+                issues.append(f"Cluster Operator {co_name} is not Available")
+            if condition_type == "Upgradeable" and condition_status == "False":
+                issues.append(f"Cluster Operator {co_name} is not Upgradeable")
```
eliorerz, line 369: Maybe it worth checking also for `Progressing=True`. If an operator is still progressing from a previous operation, starting an upgrade could cause issues.

eliorerz (changes requested):

reconcile/cluster_upgrade.py
rporres: It currently works under the same assumption as cli.py, an enclave installation in which dependencies have been installed (ansible-core and ansible_pip_requirements.txt). FWIW, it is the same assumption we have with our ansible playbooks.

Having said that, I'm taking a look into moving to a standard pyproject.toml, with more modern dependencies management and a test harness so that we can add tests to cli commands. That's what I meant by "dependencies scaffolding missing" 😄 

But I can start with that part if you think it is more convenient.

reconcile/cluster_upgrade.py
rporres: thx, I'll look into that.

reconcile/cluster_upgrade.py
maorfr: let's try to have this file not require any additional dependencies in the beginning, like in #123. we want this cli to be lean and composable so it can serve multiple purposes and executions.

i believe we should revisit this approach once we have a more standard cli?

rporres: #123 was written under the assumption of using the Python 3.9 interpreter that comes in the ocp image. That's the main reason why it was written the way it was, even if 3.9 hasn't been supported in 6 months. We cannot write new software with those assumptions, honestly.

My assumption of this cli was that it runs in the same environment that we run our ansible playbooks: Python 3.12 and the set of dependencies we install to be able to run our environment as defined in the current requirements file. Why do we have to assume a different environment to run our cli? It's going to be quite messy if we set ourselves to support different execution environments.

Removed `package.version` in 44d3899e5292588b2cdabe4a7412610e4dae45ca. Let's work on the above as it is important to understand where this is going to be run.

reconcile/cluster_upgrade.py
```diff
@@ -0,0 +1,483 @@
...
+
+    if available_versions is None:
+        raise UpdateGraphUnavailableError()
+
+    logging.debug(
+        "Checking if desired version %s is in available list", desired_version
+    )
+
+    if desired_version not in available_versions:
+        raise VersionNotAvailableError(desired_version, available_versions)
```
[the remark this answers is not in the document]
rporres: not relevant anymore, we moved away from x.y or package.version

maorfr (approved):
LGTM

let's also add an entrypoint to cluster_upgrade.py that does not require even click. to enable cluster_upgrade.py to be called as a standalone script.

rporres:
> let's also add an entrypoint to cluster_upgrade.py that does not require even click. to enable cluster_upgrade.py to be called as a standalone script.

Discussed this with @maorfr offline. There's desire of being able to run this under a Tekton Pipeline. The two alternatives are: 

* Using the `ocp-release` image. That means using a Python version (3.9) that's no longer supported by the community and that's different from the version we use for the rest of the project (3.12, from RHEL10)
* Creating a container for that. That makes supporting one more artifact but keeps Python dependencies sane and uniform for the whole project (ansible + cli) which makes things easier to manage.

So for the time being we move on with the idea of a single chain of dependencies and we'll call cluster upgrade from cli.py

reconcile/cluster_upgrade.py
rporres: fixed in a70e85fa0ccdc7dcd723f3aec563d68d80fb2c1c

rporres:
@eliorerz I think your comments have been addressed. Can you check?

reconcile/cluster_upgrade.py
rporres: reverted in 6919737. Taking a look to the documentation https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html-single/updating_clusters/index, the conditions already listed should be enough to determine if a cluster can be upgraded wrt operators status and `Progressing` is not mentioned as a way to determine if moving to a different runlevel.

eliorerz (approved):

PR #346, opened 2026-05-12
enhance vast-csi plugin with infra tenant provisioning

wgordon17:
## Summary
- Promotes vast-csi from mirror-only addon to a full foundation storage plugin with operator installation, CSI driver lifecycle, per-tier StorageClass provisioning, and Quay NFS integration
- Introduces the infra tenant pattern: platform services (Quay, monitoring) get their own isolated VAST tenant with quotas, following osac-aap naming conventions so both systems can reason about each other's resources
- Adds clusterSelector for spoke cluster fleet deployment via ACM (depends on #340)


maorfr:
the slack discussion led me to understand vast is required on spoke clusters, but the PR does most of the work directly on the management cluster. how does that sit together?

i do like the implementation, and if we want to support vast as a core piece that would also make sense.

please re-submit your PR without forking the repository (granted you permissions).

wgordon17:
@maorfr submitted #349 now, also with the addition of https://github.com/rh-ecosystem-edge/enclave/pull/349/changes#diff-eb5268bf711057688c45cb5ac440709a70f308bf9a434ec94e57503878365685R52-R54 to install the CSI operator onto spoke clusters....other than that, the spoke integrations should be managed via the osac-aap integration

PR #351, opened 2026-05-14
ACM MultiClusterHub disable update ClusterImageSets

maorfr:
ref: https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.15/html/install/installing#advanced-config-hub

we manage our own ClusterImageSets: https://github.com/rh-ecosystem-edge/enclave/blob/953808a2915e4ef5dd29f3d30bd4f2b29f582a6a/operators/advanced-cluster-management/acm_cis.yaml

disable to avoid updates and possibly initial population.

## Summary by CodeRabbit

- **Configuration**
  - Disabled automatic updates to cluster image sets in the MultiClusterHub configuration.
  - Existing subscription annotation and readiness behavior remain unchanged.


maorfr:
related to https://github.com/gori-project/GoRI/issues/953

PR #322, opened 2026-04-30
extract cincinnati plugin

maorfr:
extract the cincinnati plugin to a plugin.

## Summary by CodeRabbit

* **Configuration Updates**
  * Cincinnati plugin enabled by default alongside the selected storage plugin.
  * Cincinnati operator removed from the global operator list and managed via the plugin.
  * Registry mirror for the update service moved from global registries into the plugin scope.

* **CI / Deployment**
  * End-to-end deployment workflow now deploys the Cincinnati foundation plugin during disconnected e2e runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/322)

PR #363, opened 2026-05-17
Fix missing Quay registry CA in image.config for UpdateService

danielerez:
Configure image.config.openshift.io cluster additionalTrustedCA before the Cincinnati operator reconciles UpdateService. Registry CA trust was only applied in quay_disconnected_finish, which runs after cincinnati-operator in phase 5, leaving UpdateService with RegistryCACertFound=False.

The quay-registry-ca ConfigMap must include both the registry hostname key (for node image pulls) and updateservice-registry (required by the Update Service operator). Without the CA trusted, nodes fail to pull graph/release images from the disconnected registry, upgrade reconciliation breaks, and control plane nodes can remain cordoned (Ready,SchedulingDisabled).

Extract shared trust_quay_registry_ca_for_image_config tasks, invoke them from cincinnati-operator after UpdateService creation, and wait for RegistryCACertFound=True. Reuse the same include from quay_disconnected_mirrors.yaml.


maorfr:
switching cincinnati to an addon plugin in #322 would cause it to be deployed after quay mirroring is complete

danielerez:
> switching cincinnati to an addon plugin in #322 would cause it to be deployed after quay mirroring is complete

But we'll still need to create a ConfigMap with the Quay registry CA, no? (I mean, as done in this change)
I.e. since the UpdateService looks for additionalTrustedCA in image config: https://github.com/rh-ecosystem-edge/enclave/pull/363/changes#diff-4e6dbae8084eb2d3bf9e64ad78e7a1609f1a521d0d203a02878eb3740442ae59R85

maorfr (approved):

PR #343, opened 2026-05-11
Migrate to pyproject.toml + uv for Python dependency management

rporres:
Centralizes Python dependencies in pyproject.toml with uv as the package
manager replacing the fragmented setup.

Added tests to reconcile cli.

## Summary by CodeRabbit

* **Infrastructure**
  * Switched Python dependency management to uv and bumped project Python to 3.12.
  * CI: added Python validation (lint, types, tests), artifact on failure, and image rebuild triggers when Python config changes.
  * Setup scripts and workflows updated to use the new install flow.

* **Tests**
  * Added basic unit tests
 
* **Other**
  * Updated ignore patterns for Python caches.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/343)

PR #364, opened 2026-05-17
Fix ModuleNotFoundError: No module named 'click' on operator install

danielerez:
When locally invoking 'make -f Makefile.ci ci-flow-disconnected': Operator playbook fails in 'Approve Install Plan (multicluster-engine)' since reconcile/cli.py ran on the Landing Zone without ansible_pip_requirements installed.

    Traceback (most recent call last):
      File "/home/cloud-user/enclave/playbooks/../reconcile/cli.py", line 1, in <module>
        import click
    ModuleNotFoundError: No module named 'click'

Thus, install ansible_pip_requirements.txt at the start of playbooks/05-operators.yaml so reconcile CLI and collection dependencies are present before configure_operator.

### DIFF
--- a/playbooks/05-operators.yaml
+++ b/playbooks/05-operators.yaml
@@ -25,6 +25,13 @@
           tags: always
       tags: always
 
+    - name: Ensure Python requirements for reconcile CLI and collections (ansible_pip_requirements.txt)
+      ansible.builtin.include_tasks:
+        file: tasks/ensure_reconcile_python_requirements.yaml
+        apply:
+          tags: always
+      tags: always
+
     - name: Display operator installation mode
       ansible.builtin.debug:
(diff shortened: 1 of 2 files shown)


maorfr:
i believe this will be handled in #343

danielerez:
> i believe this will be handled in #343

Good point.
@rporres Would moving to uv package mitigate this issue (ModuleNotFoundError: No module named 'click')?

rporres:
Moving to `uv` alone will not mitigate this issue directly, it just aims to unify how dependencies are installed. But https://github.com/rh-ecosystem-edge/enclave/pull/343 also fixes `install-enclave` target by properly installing dependencies using `setup_*` scripts. Since `install-enclave` is part of `ci-flow-disconnected` target it should fix your issue.

Closing this, please retest once #343 is merged.

PR #367, opened 2026-05-18
bump openshift-ai to 3.4.0

maorfr:


maorfr:
since the ACM policy that deploys OpenShift AI is set to Automatic installPlanApproval, this needs to be investigated and discussed further before merging

investigating in https://github.com/rh-ecosystem-edge/enclave/pull/454

maorfr:
replaced by #840

PR #376, opened 2026-05-18
Ensure global variables are checked via bootstrap.sh

rporres:
The schema validation playbook was never validating user config files —
passing -e@config/global.yaml only loaded variables into scope, it did
not trigger any schema check. Unknown properties like a mistyped key
were silently accepted.

Fix by loading all user config files through load-vars.yaml and validating
their merged content. Split validation into two tags: validate-config
(config files, run by bootstrap.sh)and test-fixtures (test fixtures, run by CI
only).

Since we are honouring schema properly, missing properties have been added
so that this doesn't break any current installation and we have removed
the unused blockStorageBackend alias.

VAST schema has been added to avoid changes that will potentially break
working deployments. Documentation has been left out for a subsequent PR
to avoid cluttering this patch any further.

validate schema tests now have no_log as they can leak secrets. A task
to upload redacted config files have been added to be able to inspect
the configuration if something breaks.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed deprecated blockStorageBackend; use storage_plugin.

* **New Features**
  * Added vast-csi support and required VAST fields.
  * New vars: enabled_plugins, pullSecretPath, and disconnected cache registry fields.
  * Redfish BMC addresses may include optional :port.

* **Tests**
  * Expanded/updated schema fixtures and validation coverage (valid/invalid cases).

* **Chores**
  * Consolidated schema validation flow and updated validation tags.
  * CI artifact collection now includes redacted enclave config files.

* **Documentation**
  * Documented quayBackendLocalStorageConfiguration and Redfish ip:port behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/376?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/verification/collect_ci_artifacts.sh
+++ b/scripts/verification/collect_ci_artifacts.sh
@@ -578,6 +578,45 @@ collect_lz_config_files() {
     scp $ssh_opts cloud-user@"$lz_ip":/home/cloud-user/.openshift_install.log "${OUTPUT_DIR}/landing-zone/openshift_install_agent.log" 2>/dev/null || true
 }
 
+collect_lz_enclave_config() {
+    local lz_ip="$1"
+    local ssh_opts="-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ConnectTimeout=10 -q"
+    local config_dir="${OUTPUT_DIR}/landing-zone/config"
+
+    mkdir -p "$config_dir"
+    info "Collecting enclave configuration files from Landing Zone..."
+
+    # Redact any key matching /password|secret|key|cert/i at all nesting levels before collecting.
+    # String values that are valid JSON are parsed and redacted recursively (e.g. odfExternalConfig).
+    local redact_py='
+import yaml, sys, re, json
+PATTERN = re.compile(r"password|secret|key|cert", re.IGNORECASE)
+def redact(obj):
+    if isinstance(obj, dict):
+        return {k: "REDACTED" if PATTERN.search(k) else redact(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [redact(i) for i in obj]
+    if isinstance(obj, str):
+        try:
+            return json.dumps(redact(json.loads(obj)))
+        except (json.JSONDecodeError, ValueError):
+            pass
+    return obj
+data = yaml.safe_load(open(sys.argv[1]))
+print(yaml.dump(redact(data), default_flow_style=False))
+'
+
+    for cfg in global.yaml certificates.yaml cloud_infra.yaml; do
+        local remote_path="/home/cloud-user/enclave/config/${cfg}"
+        local out
+        out=$(ssh $ssh_opts cloud-user@"$lz_ip" \
+            "[ -f ${remote_path} ] && python3 -c '${redact_py}' ${remote_path}" 2>/dev/null) \
+            && [ -n "$out" ] \
+            && echo "$out" > "${config_dir}/${cfg}" \
+            || warn "Could not collect ${cfg}"
+    done
+}
+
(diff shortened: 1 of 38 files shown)


scripts/verification/collect_ci_artifacts.sh
```diff
@@ -578,6 +578,43 @@ collect_lz_config_files() {
...
+    return obj
+data = yaml.safe_load(open(sys.argv[1]))
+print(yaml.dump(redact(data), default_flow_style=False))
+'
+
+    for cfg in global.yaml certificates.yaml cloud_infra.yaml; do
+        local remote_path="/home/cloud-user/enclave/config/${cfg}"
+        ssh $ssh_opts cloud-user@"$lz_ip" \
+            "[ -f ${remote_path} ] && python3 -c '${redact_py}' ${remote_path}" \
+            > "${config_dir}/${cfg}" 2>/dev/null \
```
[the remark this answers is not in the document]
rporres: pyyaml is installed in the LZ via ansible, we know that since we control dependencies

test-fixtures/schemas/invalid/invalid-enum.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-# Invalid fixture: unknown blockStorageBackend value
-# Expected failure: blockStorageBackend must be one of [lvms, odf].
+# Invalid fixture: unknown storage_plugin value
+# Expected failure: storage_plugin must be one of [lvms, odf].
```
rccrdpccl: nit: this comment should also include vast-csi

scripts/verification/collect_ci_artifacts.sh
```diff
@@ -578,6 +578,45 @@ collect_lz_config_files() {
...
+    if isinstance(obj, list):
+        return [redact(i) for i in obj]
+    if isinstance(obj, str):
+        try:
+            return json.dumps(redact(json.loads(obj)))
+        except (json.JSONDecodeError, ValueError):
+            pass
+    return obj
+data = yaml.safe_load(open(sys.argv[1]))
+print(yaml.dump(redact(data), default_flow_style=False))
```
rccrdpccl, line 606: another approach to redact sensitive fields is used here https://github.com/rh-ecosystem-edge/enclave/blob/main/scripts/diagnostics/gather_lz.sh#L129-L137
should we unify them and maybe getting rid of this inline python? Maybe a little python helper that can be executed from different parts of the codebase?

rporres: hmm, it really makes sense to unify this and use a less blunt approach. A helper definitely makes sense. Since this is kind of a side quest in this big PR (I only added it because I realised that the validation schema could leak credentials and I needed to be able to debug mistakes), do you mind if I take care in a subsequent patch?

test-fixtures/schemas/invalid/invalid-enum.yaml
rporres: fixed in b8d34a3ba8c86b1de4cafc31f14e70dc8f7e917c

scripts/verification/collect_ci_artifacts.sh
rccrdpccl: I do not mind of course :)  let's focus on this on a follow up PR

rccrdpccl:
LGTM

rccrdpccl (approved):

scripts/verification/collect_ci_artifacts.sh
rporres: cool, thx

maorfr (approved):

PR #381, opened 2026-05-20
OSAC-669: Apply PinnedImageSet to handle LZ registry failover

danielerez:
To ensure the mgmt cluster can recover from node reboots and crashes even without an available registry, image resilience is necessary. This can be achieved by pinning the required images using the PinnedImageSet functionality, as detailed in the OpenShift documentation on Pinning and Preloading Images [*].
This action stops CRI-O from triggering Garbage Collection for the images, and thus ensuring the the required images exist for getting the cluster back to an healthy status.

[*] https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/machine_configuration/machine-config-pin-preload-images-about#machine-config-pin-preload-images-about

## Summary by CodeRabbit

* **New Features**
  * Optional master PinnedImageSet applied after Quay mirror (can be skipped via config or during mirror dry-run).
  * Automatic per-master-node image digest collection and templating into a PinnedImageSet manifest.
  * New CLI command to collect node image digests and a template to render PinnedImageSet manifests.

* **Improvements**
  * Better per-node diagnostics, stricter failure conditions, deduplication and exclusion filtering, and clearer logging when results are empty or truncated.

* **Tests**
  * Added tests for digest parsing, exclusion rules, CLI behavior, timeout handling, and manifest rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/381?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/config/global.example.yaml
+++ b/config/global.example.yaml
@@ -120,6 +120,10 @@ storage_plugin: lvms
 # Log level for oc-mirror (default: info, options: trace, debug, info, error)
 # ocMirrorLogLevel: debug
 
+# Apply master PinnedImageSet after Quay mirror (default: true). Set false to skip
+# post-mirror prefetch pinning (e.g. faster iteration or when MCO pin step is unwanted).
+# quayPinnedImageSetEnabled: false
+
 # Additional NTP sources for cluster nodes (no additional servers by default)
 # defaultNtpServers:
 #  - YOUR_NTP_SERVER_1
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -322,6 +322,23 @@ defaultNtpServers:
 - Only needed when cluster nodes cannot reach the default public NTP pool
 - Useful in air-gapped or firewalled environments
 
+#### `quayPinnedImageSetEnabled`
+
+**Description**: Controls whether the post-mirror master `PinnedImageSet` is generated and applied after Quay image mirroring.
+
+**Type**: Boolean (optional)
+
+**Default**: `true`
+
+**Example**:
+```yaml
+quayPinnedImageSetEnabled: false
+```
+
+**Notes**:
+- Set to `false` when you want to skip post-mirror prefetch pinning.
+- This setting is ignored when `mirror_dry_run` is enabled.
+
 ### Hardware Configuration
 
 #### `agent_hosts`
@@ -1349,6 +1366,7 @@ lzBmcIP: 100.64.1.10
 # masterMaxPods: 250 # Default: 500
 # diskEncryption: true  # Default: false (set to true to enable TPM v2 encryption)
 # ocMirrorLogLevel: debug  # Default: info
+# quayPinnedImageSetEnabled: false  # Default: true (set false to skip post-mirror pinning)
 # defaultNtpServers:  # No additional servers by default
 #  - YOUR_NTP_SERVER_1
 #  - YOUR_NTP_SERVER_2
--- a/operators/quay-operator/quay_disconnected_finish.yaml
+++ b/operators/quay-operator/quay_disconnected_finish.yaml
@@ -1,6 +1,7 @@
 ---
-# Quay disconnected tasks: imageset for oc-mirror and mirror run.
-# This file is included only when disconnected | default(true).
+# Quay disconnected finish: wait for oc-mirror, apply IDMS/ITMS and registry CA trust,
+# then optionally generate/apply master PinnedImageSet (post-mirror so prefetch trusts Quay).
+# PinnedImageSet is skipped when mirror_dry_run or quayPinnedImageSetEnabled is false.
 
 - name: Wait for async oc-mirror to quay
   ansible.builtin.include_tasks: "../../playbooks/tasks/background_task_wait.yaml"
@@ -12,14 +13,17 @@
     path: "{{ lookup('env','HOME') }}/.config/containers/registries.conf"
     state: absent
 
-- name: Apply release signature ConfigMap
-  ansible.builtin.command:
-    cmd: |
-      {{ workingDir }}/bin/oc apply -f {{ workingDir }}/config/oc-mirror-workspace-quay/working-dir/cluster-resources/signature-configmap.yaml
-  environment:
-    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
-
 # Apply oc-mirror-generated IDMS/ITMS to the cluster (fallback to internal Quay when LZ fails)
 # and add the Quay registry CA to image.config so nodes trust the registry for image pulls.
 - name: Apply Quay mirrors and trust registry CA for image pulls
   ansible.builtin.include_tasks: quay_disconnected_mirrors.yaml
+
+- name: Generate and apply master PinnedImageSet from images on master nodes
+  ansible.builtin.include_tasks: quay_pinnedimageset.yaml
+  when: >-
+    (
+      quayPinnedImageSetEnabled
+      | default(quay_pinnedimageset_enabled | default(true))
+      | bool
+    )
+    and not (mirror_dry_run | default(false) | bool)
--- a/playbooks/tasks/collect_master_node_image_digests.yaml
+++ b/playbooks/tasks/collect_master_node_image_digests.yaml
@@ -1,0 +1,102 @@
+---
+# collect_master_node_image_digests.yaml — digest pull specs from images on all master nodes.
+#
+# Sets:
+#   pinned_images — unique digest pull specs (PinnedImageSet API format).
+
+- name: List master nodes for image collection
+  kubernetes.core.k8s_info:
+    kind: Node
+    label_selectors:
+      - node-role.kubernetes.io/master
+  environment:
+    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+  register: r_pinned_master_nodes
+
+- name: Fail when no master nodes found for image collection
+  ansible.builtin.fail:
+    msg: "No master nodes matched node-role.kubernetes.io/master; cannot collect images for PinnedImageSet."
+  when: r_pinned_master_nodes.resources | default([]) | length == 0
+
+# Per-node collection via enclave tools (see tools/node_image_digests.py).
+- name: Collect container image digest refs from each master node
+  ansible.builtin.command:
+    argv:
+      - enclave
+      - tools
+      - collect-node-image-digests
+      - --node
+      - "{{ item.metadata.name }}"
+      - --oc
+      - "{{ workingDir }}/bin/oc"
+      - --raw-output-file
+      - "{{ workingDir }}/config/pinned-image-digests-{{ item.metadata.name }}.log"
+  environment:
+    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+  loop: "{{ r_pinned_master_nodes.resources }}"
+  register: r_master_crictl_digests
+  changed_when: false
+  ignore_errors: true
+  loop_control:
+    label: "{{ item.metadata.name }}"
+
+- name: Build pinned image list from master node crictl output
+  ansible.builtin.set_fact:
+    pinned_images: >-
+      {{
+        (r_master_crictl_digests.results | default([])
+        | map(attribute='stdout_lines')
+        | flatten
+        | map('trim')
+        | select()
+        | unique
+        | list)
+      }}
+
+- name: Identify failed or empty master node image collection results
+  ansible.builtin.set_fact:
+    _pinned_collection_failed_nodes: >-
+      {{
+        r_master_crictl_digests.results | default([])
+        | selectattr('failed', 'equalto', true)
+        | list
+      }}
+    _pinned_collection_empty_nodes: >-
+      {{
+        r_master_crictl_digests.results | default([])
+        | rejectattr('failed', 'equalto', true)
+        | selectattr('stdout_lines', 'defined')
+        | selectattr('stdout_lines', 'equalto', [])
+        | list
+      }}
+    _pinned_collection_has_issues: >-
+      {{
+        (r_master_crictl_digests.results | default([]) | selectattr('failed', 'equalto', true) | list | length > 0)
+        or (
+          r_master_crictl_digests.results | default([])
+          | rejectattr('failed', 'equalto', true)
+          | selectattr('stdout_lines', 'defined')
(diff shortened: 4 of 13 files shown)


maorfr:
can we obtain the list of images using `oc adm release info`?

danielerez:
> can we obtain the list of images using `oc adm release info`?

Yeah, that was my first attempt:)
But from my testings, it's problematic to pin all images in the release.
Besides taking forever to pin numerous images, the main issue is failure in specific images.
I.e. images with single manifests are failing on MCO prefetch (as it's using 'podman manifest inspect - which exits with error 125). So I've added them to an exclude list: https://github.com/rh-ecosystem-edge/enclave/pull/381/changes#diff-732cb47bb018901457d4500d498c66d1e25bc0995845f2b9de84b4b454119d39R127
The alternative done here is to simply pin all the images already pulled by the master nodes. Thus, we can ensure the prefetching process by MCO when pinning images would be quick and successful (and avoid caching uneeded images).

playbooks/tasks/collect_master_node_image_digests.yaml
```diff
@@ -0,0 +1,205 @@
...
+    state: directory
+
+# crictl images JSON from each master. Inline Python: find JSON array or {"images":[…]} (oc noise),
+# normalize @sha256 hex case, keep refs matching PinnedImageSet API (FQDN-style registry host).
+#
+# When crictl lists multiple repoDigests per image id, keep the last digest (MCO prefetch uses podman manifest inspect).
+# Default excludes skip openshift-pipelines/*, /openshift/graph-image (Cincinnati graph-data image),
+# /redhat/community-operator-index, /redhat/certified-operator-index, /redhat/redhat-marketplace-index,
+# /redhat/redhat-operator-index, /rhel9/support-tools (single-manifest refs → exit 125).
+- name: Collect container image digest refs from each master node
```
rporres, line 22: inline scripts are complicated to debug and maintain. Although it is not exactly a reconcile like command, we can add it as a reconcile command for the moment and think about we want to organize the whole enclave-cli afterwards. We already have a cli command that is called from a playbook, see https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/configure_operator.yaml#L146

danielerez: > inline scripts are complicated to debug and maintain. Although it is not exactly a reconcile like command, we can add it as a reconcile command for the moment and think about we want to organize the whole enclave-cli afterwards. We already have a cli command that is called from a playbook, see https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/configure_operator.yaml#L146

Right, good point!
Using cli commands is a nice idea. Tried to migrate the logic to a new cmd, hope it looks better:)

reconcile/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+# upstream OLM catalog indexes under /redhat/*-operator-index, disconnected mirrors of
+# those catalogs (mirror-redhat-operators, mirror-redhat-operators-<plugin>), and
+# /rhel9/support-tools (single-manifest refs that can cause prefetch exit 125).
+DEFAULT_PINNED_IMAGE_EXCLUDE_CONTAINS: tuple[str, ...] = (
+    "openshift-pipelines/",
+    "/openshift/graph-image",
+    "/redhat/community-operator-index",
+    "/redhat/certified-operator-index",
+    "/redhat/redhat-marketplace-index",
+    "/redhat/redhat-marketplace-index",
```
rporres: duplicate

config/global.example.yaml
```diff
@@ -147,6 +147,10 @@ storage_plugin: lvms
 # Log level for oc-mirror (default: info, options: trace, debug, info, error)
 # ocMirrorLogLevel: debug
 
+# Apply master PinnedImageSet after Quay mirror (default: true). Set false to skip
+# post-mirror prefetch pinning (e.g. faster iteration or when MCO pin step is unwanted).
+# quay_pinnedimageset_enabled: false
```
rporres: you will need to add this property to the global schema and document it.

nit: Although we're not strict, the configuration favours camel case.

src/tools/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+)
+_DIGEST_REF = re.compile(r"^(.+)@(sha256:[a-fA-F0-9]{64})$")
+
+# Substring filters for digest refs to skip when building a PinnedImageSet list.
+# When crictl lists multiple repoDigests per image id, MCO prefetch runs
+# `podman manifest inspect`; these defaults avoid images that break or are not
+# worth pinning: openshift-pipelines/*, Cincinnati graph-data (/openshift/graph-image),
+# upstream OLM catalog indexes under /redhat/*-operator-index, disconnected mirrors of
+# those catalogs (mirror-redhat-operators, mirror-redhat-operators-<plugin>), and
+# /rhel9/support-tools (single-manifest refs that can cause prefetch exit 125).
```
rporres, line 29: I think the documentation deserves a discussion on the tradeoffs we're accepting when adding this feature. It looks we're targeting etcd, kube-apiserver, OVN, MCO but some operators will not be functional.

reconcile/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+    logger.debug(
+        "node=%s collected %d digest ref(s) from %d byte(s) of crictl output",
+        node,
+        len(refs),
+        len(raw_output),
+    )
+    return CollectResult(refs=refs, raw_output=raw_output)
+
+
+def emit_collection_output(result: CollectResult) -> None:
```
rporres: docstrings on all public methods would be appreciated...

src/tools/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+    for value in values:
+        if value not in unique:
+            unique.append(value)
+    return unique
+
+
+def _select_repo_digests(rd_vals: list[str]) -> list[str]:
+    if not rd_vals:
+        return []
+    return rd_vals if len(rd_vals) == 1 else [rd_vals[-1]]
```
rporres, line 132: why keeping just the last one? I guess this deserves a comment 😄

src/tools/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+from dataclasses import dataclass
+
+logger = logging.getLogger(__name__)
+
+OC_DEBUG_CRICTL_TIMEOUT_SECONDS = 600
+
+# FQDN-style registry host (PinnedImageSet API).
+_OCI_NAME = re.compile(
+    r"^([a-zA-Z0-9-]+\.)+[a-zA-Z0-9-]+(:[0-9]{2,5})?/([a-zA-Z0-9-_]{0,61}/)?[a-zA-Z0-9-_.]*$"
+)
```
rporres, line 20: should we support images like `registry.example.com/org/sub-org/image` or at least log that we're discarding them when using this regex?

operators/quay-operator/quay_disconnected_finish.yaml
```diff
@@ -12,14 +13,13 @@
     path: "{{ lookup('env','HOME') }}/.config/containers/registries.conf"
     state: absent
 
-- name: Apply release signature ConfigMap
-  ansible.builtin.command:
-    cmd: |
-      {{ workingDir }}/bin/oc apply -f {{ workingDir }}/config/oc-mirror-workspace-quay/working-dir/cluster-resources/signature-configmap.yaml
-  environment:
-    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
-
```
rporres, line 21: can you add a note in the description about why this is removed?

playbooks/tasks/collect_master_node_image_digests.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - /redhat/community-operator-index
+      - /redhat/certified-operator-index
+      - /redhat/redhat-marketplace-index
+      - /redhat/redhat-operator-index
+      - mirror-redhat-operators
+      - /rhel9/support-tools
+  environment:
+    KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+    PINNED_IMAGE_EXCLUDE_CONTAINS: >-
+      {{ pinned_image_exclude_contains | default(_pinned_image_exclude_contains_default) | to_json }}
```
rporres: is this going to be called outside this task so that it's worth the list duplication in `PINNED_IMAGE_EXCLUDE_CONTAINS` and `DEFAULT_PINNED_IMAGE_EXCLUDE_CONTAINS`. Shouldn't we keep just one so that it's clear which one we should maintain?

reconcile/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
...
+    if not result.refs:
+        sys.stderr.write(result.raw_output)
```
rporres: how big is this? Should we log into a file an reference to it?

config/global.example.yaml
danielerez: Right, good points. Added to schema/doc and modified the naming.

operators/quay-operator/quay_disconnected_finish.yaml
danielerez: Oh, it was just a cleanup, as it's handled now by another task in configure_ocp_abi: https://github.com/rh-ecosystem-edge/enclave/blob/b6d6658963adea61d41230d37266dc869dbba38b/playbooks/tasks/configure_ocp_abi.yaml#L72
Anyway, it's already removed in master:)
https://github.com/rh-ecosystem-edge/enclave/blob/main/operators/quay-operator/quay_disconnected_finish.yaml

src/tools/node_image_digests.py
```diff
@@ -0,0 +1,242 @@
+"""Collect PinnedImageSet-format digest refs from crictl images JSON on a node."""
```
rporres, line 1: following our conversations, python code has been reorganized so that we have an `enclave-tools` command for the non-reconcile python tooling. Please move this script there. See https://github.com/rh-ecosystem-edge/enclave/pull/418

playbooks/tasks/collect_master_node_image_digests.yaml
danielerez: Good catch! Yeah, I indeed considered to support calling from outside.
But not crucial for now, removed the duplication to make it cleaner.

src/tools/node_image_digests.py
danielerez: Sure, I'll modify the regex to support both.

reconcile/node_image_digests.py
danielerez: indeed, thanks:)

src/tools/node_image_digests.py
danielerez: So this is addressing only the images we have to exclude to avoid pull issues.
The approach is to basically pin all images from masters nodes.
I.e. We basically want to ensure that quay operator could run post nodes crash. So the internal quay registry would serve any required images needed for the cluster. Thus allowing the cluster recovering from master nodes restarts.

src/tools/node_image_digests.py
danielerez: Right, will add:)

reconcile/node_image_digests.py
danielerez: Yeah, it could indeed be an issue.
I'll output to a log file instead.

reconcile/node_image_digests.py
danielerez: Sure, sorry, will do.

operators/quay-operator/quay_disconnected_finish.yaml
rporres: fair enough :)

docs/CONFIGURATION_REFERENCE.md
```diff
@@ -284,6 +284,24 @@ defaultNtpServers:
...
+
+**Example**:
+```yaml
+quayPinnedImageSetEnabled: false
+```
+
+**Notes**:
+- Set to `false` when you want to skip post-mirror prefetch pinning (for example, to speed up iteration loops)
+- This setting is ignored when `mirror_dry_run` is enabled
+- Legacy alias `quay_pinnedimageset_enabled` is still accepted for backward compatibility
```
rporres: no, it is not accepted, the schema validation is strict and won't let this pass in the global config.

src/tools/node_image_digests.py
```diff
@@ -0,0 +1,274 @@
...
+    output to that path and prints a short message to stderr instead of dumping it.
+    """
+    if not result.refs:
+        if raw_output_file:
+            output_path = Path(raw_output_file)
+            output_path.parent.mkdir(parents=True, exist_ok=True)
+            output_path.write_text(result.raw_output, encoding="utf-8")
+            sys.stderr.write(
+                f"No digest refs were collected; raw output saved to {output_path}\n"
+            )
```
rporres, line 257: logger.error?

docs/CONFIGURATION_REFERENCE.md
danielerez: Right, removed.

src/tools/node_image_digests.py
danielerez: Actually enclave-tools currently does not call logging.basicConfig. I.e. as opposed to enclave-reconcile. So it would use a different format if replacing...

rporres: hmmm, that's a mistake from my side... it definitely should. It will get fixed.

rporres (approved):

playbooks/tasks/collect_master_node_image_digests.yaml
```diff
@@ -0,0 +1,101 @@
...
+- name: Fail when no master nodes found for image collection
+  ansible.builtin.fail:
+    msg: "No master nodes matched node-role.kubernetes.io/master; cannot collect images for PinnedImageSet."
+  when: r_pinned_master_nodes.resources | default([]) | length == 0
+
+# Per-node collection via enclave-tools (see tools/node_image_digests.py).
+- name: Collect container image digest refs from each master node
+  ansible.builtin.command:
+    argv:
+      - enclave-tools
```
rporres: This got separated in https://github.com/rh-ecosystem-edge/enclave/pull/425

It is now `enclave tools`

danielerez: Nice, thanks. Rebased and fixed.

rporres (approved):

PR #386, opened 2026-05-21
Add CNV plugin for OpenShift Virtualization deployment

eurijon:
Installs the kubevirt-hyperconverged operator (v4.20.14, stable channel) and deploys the HyperConverged CR.

Removes kubevirt-hyperconverged from metallb extraMirrorPackages since mirroring is now handled by the plugin.

OSAC-930

## Summary by CodeRabbit

* **New Features**
  * Added CNV virtualization plugin with automated operator deployment, resource configuration, and health status verification.

### DIFF
--- a/defaults/operators.yaml
+++ b/defaults/operators.yaml
@@ -111,5 +111,3 @@ operators:
     global: true
     csvNames:
       - metallb-operator
-    extraMirrorPackages:
-      - kubevirt-hyperconverged
--- a/defaults/operators.yaml
+++ b/defaults/operators.yaml
@@ -111,3 +111,5 @@ operators:
     global: true
     csvNames:
       - metallb-operator
+    extraMirrorPackages:
+      - kubevirt-hyperconverged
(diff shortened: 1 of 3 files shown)


defaults/operators.yaml
```diff
@@ -111,5 +111,3 @@ operators:
     global: true
     csvNames:
       - metallb-operator
-    extraMirrorPackages:
-      - kubevirt-hyperconverged
```
maorfr, line 115: this will impact existing enclave installations

eurijon: ok, it's still not clear to me if mirroring `kubevirt-hyperconverged` is required for an enclave base installation. If we have a plugin for kubevirt, doesn't make sense to handle images from it?

maorfr: it makes perfect sense, but this package should be included by default for backwards compatibility.

eurijon: sure, `kubevirt-hyperconverged` entry is restored  in https://github.com/rh-ecosystem-edge/enclave/pull/386/changes/8b11caeb39b7fac683279b3c356c06f375428350

maorfr (approved):
LGTM

note there might be a conflict in the mirroring/catalogs of kubevirt. not a blocker.

eurijon:
> LGTM
> 
> note there might be a conflict in the mirroring/catalogs of kubevirt. not a blocker.

ack, will consider it when adding the support for disconnected deployment

PR #388, opened 2026-05-21
OSAC-955: Add generic Helm image extraction for plugin mirroring

eliorerz:
## Summary

- Add opt-in `extractImages: true` flag on `helm[]` entries in the plugin schema
- Before mirroring, run `helm template` on qualifying local charts using chart defaults to discover container image references
- Merge extracted images into `plugin.additionalImages` (deduplicated), eliminating the need to manually hardcode images already declared in chart defaults
- Local charts only — entries with `repo` set are skipped; charts without `extractImages` are unaffected

## Test plan

- [ ] `scripts/verification/validate_plugins.sh` passes for all plugins
- [ ] Schema validation accepts `extractImages: true` on helm entries
- [ ] Test with `plugins/helm-test`: add `extractImages: true`, verify chart image is extracted and merged into additionalImages
- [ ] Verify remote charts (with `repo:`) are skipped
- [ ] Verify charts without `extractImages` are skipped
- [ ] Full disconnected deploy test with a plugin using `extractImages: true`

## Summary by CodeRabbit

* **New Features**
  * Automatic extraction of container images from local Helm charts when enabled via the chart's extractImages flag; extracted images are merged into the mirroring list.
* **Documentation**
  * Plugin docs and lifecycle steps updated to describe the extractImages option and the new "Extract Helm images" step (local charts only).

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/388?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/PLUGIN_ARCHITECTURE.md
+++ b/docs/PLUGIN_ARCHITECTURE.md
@@ -75,7 +75,7 @@ registries:
 | `additionalImages` | Extra images to include in the plugin's oc-mirror image set. |
 | `blockedImages` | Images to exclude from mirroring (by tag, digest, or pattern). |
 | `requires` | Declarative requirements validated at load time, before any deployment work begins. See [Load-Time Validation](#load-time-validation). |
-| `helm` | List of Helm charts to install after operators, before `tasks/deploy.yaml`. Supports local charts and remote repos. Each entry specifies `release`, `namespace`, and optional `repo`, `version`, and values template. |
+| `helm` | List of Helm charts to install after operators, before `tasks/deploy.yaml`. Supports local charts and remote repos. Each entry specifies `release`, `namespace`, and optional `repo`, `version`, values template, and `extractImages`. When `extractImages: true` is set on a local chart, `helm template` is run before mirroring to discover container image references and merge them into `additionalImages` automatically. |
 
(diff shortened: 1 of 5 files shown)


maorfr (approved):
LGTM

why is this only implemented for local charts?
fwiw, if we use `helm template` we might as well use `oc apply` instead of `helm install / upgrade`.

eliorerz:
@maorfr 


> Why only local charts?

It's a practical constraint around pipeline timing. The extraction runs before mirroring, and at that point remote charts haven't been fetched yet. You'd need to helm repo add + helm pull first just to template them. It's doable but adds complexity for a case that doesn't exist yet (no plugin currently uses remote charts with extractImages). If a need comes up, it can be extended.

> If we use helm template we might as well use oc apply

I hope I understood your question but I think this comes from a misread of what the code does. Helm template is only used at mirror time to discover image references the output is parsed for `image:` lines and then thrown away. The actual deployment still goes through `helm upgrade --install` as before. We're not replacing Helm's release management (rollback, upgrade tracking, helm status, release history) - we're just peeking at the chart's default values to figure out which images need mirroring.

PR #395, opened 2026-05-25
Fix operator reconcile output not showing in Ansible logs

eurijon:
## Summary
- The `enclave-reconcile` script uses Python's `logging` module, which writes to stderr. The Ansible debug task was printing `stdout_lines`, which was always empty (`[]`). Changed to `stderr_lines` so the install plan approval output is actually visible in Ansible logs.
- Improved the task name from generic "Print the output" to include the operator name, consistent with other tasks in the file.

## Test plan
- [ ] Run an Enclave install and verify the "Print Install Plan approval output" task shows the reconcile log lines instead of an empty list

## Release Notes

* **Bug Fixes**
  * Corrected operator configuration logging to capture output from the appropriate source stream, ensuring error messages are properly displayed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/395?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -146,12 +146,13 @@
     enclave-reconcile operator-versions --operators "{{ [operator] }}" --no-dry-run
   register: operator_update_result
 
-- name: Print the output
+- name: "Print Install Plan approval output ({{ operator.name }})"
   when:
     - r_sub_check.resources | length == 0
     - operator.namespace != "openshift-operators"
   ansible.builtin.debug:
-    var: operator_update_result.stdout_lines
+    # enclave-reconcile uses Python logging, which writes to stderr
+    var: operator_update_result.stderr_lines
 
 - name: "Get Installed CSV ({{ operator.name }})"
   when:


playbooks/tasks/configure_operator.yaml
```diff
@@ -146,12 +146,13 @@
...
 
-- name: Print the output
+- name: "Print Install Plan approval output ({{ operator.name }})"
   when:
     - r_sub_check.resources | length == 0
     - operator.namespace != "openshift-operators"
   ansible.builtin.debug:
-    var: operator_update_result.stdout_lines
+    # enclave-reconcile uses Python logging, which writes to stderr
+    var: operator_update_result.stderr_lines
```
rccrdpccl, line 155: wouldn't it be best to change the reconciler to use stdout instead?
Sending logs to stderr makes sense in a cli application that could output an actual result, but in case of a reconciler I think it would be best changing where the output is sent

rporres: I think it makes sense what @rccrdpccl says. I can take care, @eurijon, if you prefer. I'm already working on reconcile cli

eurijon: Sure, adding `logging.basicConfig(stream=sys.stdout, level=logging.INFO)` to the script would make it. I'll let it to you @rporres if you have another approach

PR #396, opened 2026-05-25
Fix OSUS graph-builder TLS trust for disconnected Quay registry

danielerez:
Cincinnati graph-builder failed to scrape release metadata from the Quay route with certificate verify failed (self signed certificate in chain) because the updateservice-registry CA in image.config was missing[1] or did not match the certificate actually presented by the registry. Thus, 'update-service' pods fail to start[2].

Apply Quay registry CA trust before creating UpdateService, resolve the correct CA PEM via enclave-reconcile resolve-quay-registry-ca (router-ca when it verifies the live leaf, otherwise the chain root) Then ensure the openshift-config ConfigMap includes updateservice-registry, and restart the OSUS deployment so graph-builder reloads trust.

[1]
```
cincinnati::plugins::internal::graph_builder::commons] unable to process
certificate ca-bundle.trust.crt: builder error: error:0909006C:PEM
routines:get_name:no start line:crypto/pem/pem_lib.c:745:Expecting: CERTIFICATE
```

[2]
```
$ oc -n openshift-update-service get pod
NAME                                      READY   STATUS                   RESTARTS         AGE
graph-data-tag-digest                     0/1     ContainerStatusUnknown   1                50m
service-5785cb4d7c-q4hdf                  0/2     Running                  24 (57s ago)     131m
service-5785cb4d7c-vfffw                  0/2     Running                  23 (3m27s ago)   131m
service-bd8f457f8-h4j2p                   0/2     Running                  36 (57s ago)     3h35m
updateservice-operator-787f968c77-8nwtf   1/1     Running                  0                10m
```

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to resolve the Quay registry CA certificate.

* **Improvements**
  * Registry CA trust is applied earlier in the UpdateService workflow.
  * Cluster image CA wiring now updates or sets the trusted CA config as needed.
  * UpdateService deployment is restarted to pick up updated CA trust.
  * CA retrieval uses a dedicated resolution step for registry routes.

* **Tests**
  * Added CLI and CA-resolution unit tests covering success, failures, timeouts, and edge cases.

### DIFF
--- a/reconcile/quay_registry_ca.py
+++ b/reconcile/quay_registry_ca.py
@@ -1,0 +1,133 @@
+"""Resolve the CA PEM used to trust the Quay registry route for OSUS and image pulls."""
+
+from __future__ import annotations
+
+import base64
+import binascii
+import logging
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_TLS_CONNECT_TIMEOUT_SECONDS = 60
+_OC_COMMAND_TIMEOUT_SECONDS = 30
+_OPENSSL_VERIFY_TIMEOUT_SECONDS = 10
+_PEM_BLOCK = re.compile(
+    r"-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----",
+    re.DOTALL,
+)
+
+
+def pem_blocks(pem_text: str) -> list[str]:
+    return _PEM_BLOCK.findall(pem_text or "")
+
+
+def get_router_ca_pem(*, oc: str = "oc") -> str:
+    try:
+        result = subprocess.run(
+            [
+                oc,
+                "get",
+                "secret",
+                "router-ca",
+                "-n",
+                "openshift-ingress-operator",
+                "-o",
+                "jsonpath={.data.tls\\.crt}",
+            ],
+            capture_output=True,
+            text=True,
+            check=False,
+            timeout=_OC_COMMAND_TIMEOUT_SECONDS,
+        )
+    except subprocess.TimeoutExpired as exc:
+        msg = "oc get secret router-ca timed out"
+        raise RuntimeError(msg) from exc
+    except OSError as exc:
+        msg = f"{oc} not found"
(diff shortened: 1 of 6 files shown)


maorfr (approved):
LGTM

a few comments which can be addressed in a follow up PR:
- shouldn't we be able to use `enclave-reconcile` directly and not using uv?
- split encalve-reconcile to something like `enclave reconcile` and `enclave utils`

danielerez:
> LGTM
> 
> a few comments which can be addressed in a follow up PR:
> 
> * shouldn't we be able to use `enclave-reconcile` directly and not using uv?

Good catch, it should be an easy fix, I can just handle it in this change.

> * split encalve-reconcile to something like `enclave reconcile` and `enclave utils`

Right, but indeed sounds like something worth a separate PR.

rporres:
happy to take a look to reorganising Python code, I already had it mind. Created https://redhat.atlassian.net/browse/OSAC-1047 for it

reconcile/quay_registry_ca.py
```diff
@@ -0,0 +1,128 @@
...
+        msg = f"unable to resolve registry CA for {hostname}"
+        raise RuntimeError(msg)
+
+    leaf = chain[0]
+    if router_ca and _openssl_verify(router_ca, leaf):
+        logger.debug("using router-ca to trust %s", hostname)
+        return f"{router_ca}\n"
+
+    logger.debug("using TLS chain root to trust %s", hostname)
+    return f"{chain[-1]}\n"
```
rporres, line 129: what if we have intermediates in the chain? Shouldn't they be returned as well?

danielerez: > what if we have intermediates in the chain? Shouldn't they be returned as well?

So the router cert should be enough here. Since for [additionalTrustedCA](https://github.com/rh-ecosystem-edge/enclave/blob/122acf54bf110a3e23991c4fec536919dce2d33d/playbooks/tasks/trust_quay_registry_ca_for_image_config.yaml#L84), we only need the CA clients must trust, not the whole chain.

rporres (approved):

PR #385, opened 2026-05-21
only mirror foundation plugins in core

maorfr:
with this change, only foundation plugins in `enabled_plugins` are mirrored as part of the core mirror.

addon plugins will be mirrored as part of their plugin deployment.

this is a behavior adjustment following #357

## Summary by CodeRabbit

* **Chores**
  * Refined plugin deployment process with improved registry collection and operator configuration handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/385?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #399, opened 2026-05-26
support certified catalog for foundation plugins

maorfr:

follows up on #385

to be able to use vast-csi as a storage plugin for quay it needs to be a foundation plugin.

this PR adds required pieces to support certified catalog as a foundation plugin.

## Release Notes

* **New Features**
  * Added support for mirroring certified Red Hat operators separately with distinct catalog identification in disconnected environments.

* **Changes**
  * VAST-CSI plugin reclassified from addon to foundation plugin.
  * Enhanced operator catalog configuration to better distinguish between standard and certified operator sources during registry mirroring.

### DIFF
--- a/playbooks/tasks/collect_core_plugin_operators.yaml
+++ b/playbooks/tasks/collect_core_plugin_operators.yaml
@@ -31,11 +31,13 @@
          | selectattr('name', 'in', enabled_plugins)
          | list }}
   ansible.builtin.set_fact:
-    _core_plugin_operators: >-
-      {{ _foundation
-         | selectattr('operators', 'defined')
-         | items2dict(key_name='name', value_name='operators') }}
+    _core_plugin_operators: |-
+      {% set tmp = {} %}
+      {% for p in _foundation if p.operators is defined %}
+      {% set _ = tmp.update({p.name: {'operators': p.operators, 'catalog': p.catalog | default('redhat')}}) %}
+      {% endfor %}
+      {{ tmp }}
 
 - name: Display plugin operators to mirror
   ansible.builtin.debug:
-    msg: "Plugin operators to include in imageset: {{ _core_plugin_operators.values() | flatten | map(attribute='name') | list }}"
+    msg: "Plugin operators to include in imageset: {{ _core_plugin_operators.values() | map(attribute='operators') | flatten | map(attribute='name') | list }}"
--- a/playbooks/tasks/template_validation/test_imagesetconfiguration.yaml
+++ b/playbooks/tasks/template_validation/test_imagesetconfiguration.yaml
@@ -4,8 +4,8 @@
 
 - name: Render imagesetconfiguration.yaml for each storage backend
   vars:
-    _core_plugin_operators: "{{ {item.name: item.operators} }}"
-    catalog_operators: "{{ {'core': operators} | combine(_core_plugin_operators) }}"
+    _core_plugin_operators: "{{ {item.name: {'operators': item.operators, 'catalog': item.catalog | default('redhat')}} }}"
+    catalog_operators: "{{ {'core': {'operators': operators, 'catalog': 'redhat'}} | combine(_core_plugin_operators) }}"
   ansible.builtin.template:
     src: ../../templates/imagesetconfiguration.yaml.j2
     dest: "{{ temp_dir.path }}/imagesetconfiguration-{{ item.name }}.yaml"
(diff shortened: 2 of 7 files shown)


templates/imagesetconfiguration.yaml.j2
```diff
@@ -26,16 +26,19 @@ mirror:
...
+{% elif c_catalog_operators.catalog == 'certified' %}
+    - catalog: {{ certified_operator_catalog }}-{{ c_name }}:{{ certified_operator_catalog_version }}
+      targetCatalog: redhat/{{ mirror_certified_rh_operator_catalog }}-{{ c_name }}
 {% else %}
     - catalog: {{ rh_operator_catalog }}-{{ c_name }}:{{ rh_operator_catalog_version }}
       targetCatalog: redhat/{{ mirror_rh_operator_catalog }}-{{ c_name }}
 {% endif %}
       packages:
-{% for operator in c_operators %}
+{% for operator in c_catalog_operators.operators %}
```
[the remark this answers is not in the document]
maorfr: this is the current code:
```
catalog_operators: "{{ {'core': {'operators': operators, 'catalog': 'redhat'}} | combine(_core_plugin_operators | default({})) }}"
```

operators/openshift-pipelines-operator-rh/operator_update_pipeline.yaml
```diff
@@ -100,8 +100,7 @@
             script: |
                 #!/bin/bash
 
-                python ./reconcile_operator_versions.py "{{ (_core_plugin_operators | default([])) + operators }}" "$(params.dry-run)" > "$(results.status-report.path)"
-                rc=$?
+                python ./reconcile_operator_versions.py "{{ ((_core_plugin_operators | default({})).values() | map(attribute='operators') | flatten | list) + operators }}" "$(params.dry-run)" > "$(results.status-report.path)"                rc=$?
```
rporres: This change doesn't belong here. We have to fix the pipeline as it is now broken, but outside this PR.

playbooks/tasks/collect_core_plugin_operators.yaml
```diff
@@ -31,11 +31,13 @@
          | selectattr('name', 'in', enabled_plugins)
          | list }}
   ansible.builtin.set_fact:
-    _core_plugin_operators: >-
-      {{ _foundation
-         | selectattr('operators', 'defined')
-         | items2dict(key_name='name', value_name='operators') }}
+    _core_plugin_operators: |-
+      {% set tmp = {} %}
+      {% for p in _foundation %}
```
rporres: how about

```suggestion
      {% for p in _foundation if p.operators is defined %}
```

to keep the old behaviour of applying this only if `operators` is defined in the enabled plugins?

operators/openshift-pipelines-operator-rh/operator_update_pipeline.yaml
maorfr: not sure i understand @rporres

rporres: The script is broken because:

* the interface of `operator_versions.reconcile` has changed
* the script cannot be run as a standalone anymore.
* We would need to check the python version of the oc cli container to see if it works with the script

The change is also broken since `$rc` is now in the same line as the invocation

The fix for this is to create a dedicated container for the reconcile cli so that we can run it. I think it doesn't belong to this PR, you can create a ticket for it and we can take care at some point.

maorfr: left this change out of the PR

playbooks/tasks/template_validation/test_imagesetconfiguration.yaml
```diff
@@ -4,8 +4,8 @@
 
 - name: Render imagesetconfiguration.yaml for each storage backend
   vars:
-    _core_plugin_operators: "{{ {item.name: item.operators} }}"
-    catalog_operators: "{{ {'core': operators} | combine(_core_plugin_operators) }}"
+    _core_plugin_operators: "{{ {item.name: {'operators': item.operators, 'catalog': item.catalog | default('redhat')}} }}"
+    catalog_operators: "{{ {'core': {'operators': operators, 'catalog': 'redhat'}} | combine(_core_plugin_operators) }}"
```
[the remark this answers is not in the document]
maorfr: it's the existing state anyway.

rporres (approved):

PR #401, opened 2026-05-26
Add sshPubKey as alternative to sshPubPath

rccrdpccl:
Accept SSH public key content directly in the config (sshPubKey field) as an alternative to providing a file path (sshPubPath). When sshPubKey is set, load-vars writes the content to a file and sets sshPubPath automatically.

Either sshPubKey or sshPubPath must be provided (anyOf in schema). This enables the wizard to accept key content from the user without needing to manage file paths in what would be a remote host.

## Release Notes

* **New Features**
  * Added support for providing an SSH public key directly in configuration, which overrides path-based references when specified
  * Updated validation to allow users to provide either an SSH public key value or a path to one

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/401?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/playbooks/common/load-vars.yaml
+++ b/playbooks/common/load-vars.yaml
@@ -14,6 +14,19 @@
   ansible.builtin.include_vars: "{{ item }}"
   loop: "{{ user_config_files }}"
 
+- name: Write SSH public key to file if content provided
+  when: sshPubKey is defined and sshPubKey | length > 0
+  block:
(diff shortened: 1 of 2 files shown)


rccrdpccl:
https://github.com/rh-ecosystem-edge/enclave/pull/402

PR #402, opened 2026-05-26
Add sshPubKey as alternative to sshPubPath

rccrdpccl:
Accept SSH public key content directly in the config (sshPubKey field) as an alternative to providing a file path (sshPubPath). When sshPubKey is set, load-vars writes the content to a file and sets sshPubPath automatically.

Either sshPubKey or sshPubPath must be provided (anyOf in schema). This enables the wizard to accept key content from the user without needing to manage file paths on what would be a remote host.

## Summary by CodeRabbit

* **New Features**
  * Support for SSH public key configuration via either inline key content or a file path; inline keys are persisted at runtime and the path is set automatically.

* **Documentation**
  * Configuration reference updated with SSH Public Key section, examples, and TOC entry.

* **Tests**
  * Added schema test fixtures covering valid inline keys and invalid/ambiguous cases.

### DIFF
--- a/playbooks/common/load-vars.yaml
+++ b/playbooks/common/load-vars.yaml
@@ -14,6 +14,25 @@
   ansible.builtin.include_vars: "{{ item }}"
   loop: "{{ user_config_files }}"
 
+- name: Write SSH public key to file if content provided
+  when: sshPubKey is defined and sshPubKey | length > 0
+  block:
+    - name: Ensure config directory exists
+      ansible.builtin.file:
+        path: "{{ workingDir }}/config"
+        state: directory
+        mode: "0750"
+
+    - name: Write sshPubKey content to file
+      ansible.builtin.copy:
+        content: "{{ sshPubKey }}\n"
+        dest: "{{ workingDir }}/config/ssh-pub-key.pub"
+        mode: "0644"
+
+    - name: Set sshPubPath from written file
+      ansible.builtin.set_fact:
+        sshPubPath: "{{ workingDir }}/config/ssh-pub-key.pub"
+
 - name: Include common configuration
   ansible.builtin.include_vars: "../../defaults/{{ config_file }}"
   loop:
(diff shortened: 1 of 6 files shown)


maorfr (approved):
LGTM

please add the proper documentation

schemas/variables.yaml
```diff
@@ -134,6 +134,9 @@ properties:
     "$ref": "#/definitions/nonEmptyString"
   rendezvousIP:
     "$ref": "#/definitions/ipv4Address"
+  sshPubKey:
+    type: string
```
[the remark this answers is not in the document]
rporres: @rccrdpccl  Please address this issue using `nonEmptyString` definition and add the new property to a fixture.

rporres (approved):

playbooks/common/load-vars.yaml
```diff
@@ -14,6 +14,25 @@
...
   loop: "{{ user_config_files }}"
 
+- name: Write SSH public key to file if content provided
+  when: sshPubKey is defined and sshPubKey | length > 0
+  block:
+    - name: Ensure config directory exists
+      ansible.builtin.file:
+        path: "{{ workingDir }}/config"
+        state: directory
+        mode: "0750"
```
eurijon, line 24: If I'm not wrong when the user sets both `sshPubKey` and `sshPubPath`, `sshPubPath` is ignored and overriden to `{{ workingDir }}/config/ssh-pub-key.pub`. In that case shouldn't user's `sshPubPath` be taken?

rccrdpccl: Yes that was the point: trying to override the sshPubPath so we don't need to pick the right var throughout the flow.
But you raise a good point, the possibility of setting both and one getting overridden is not good. I've change it to make the fields explicitly mutually exclusive

maorfr:
assuming we are good to merge here

PR #438, opened 2026-06-03
Add trust_manager_cluster_issuer variable in trust-manager plugin

eurijon:
## Summary

- Add `trust_manager_cluster_issuer` variable (defaults to `default-ca`) so downstream plugins do not depend on a hardcoded ClusterIssuer name
- Suffix the CA Secret name with `-secret` to distinguish it from the ClusterIssuer resource
- Increase ClusterIssuer wait retries to 60x15s (15 min) to accommodate slow CertificateRequest approval on Red Hat cert-manager operator
- Update defaults, schema, template, deploy task, and test fixtures

## Test plan

- [x] `make -f Makefile.ci validate-plugins` passes
- [x] `make -f Makefile.ci validate-yaml` passes
- [x] Schema test fixtures: valid base passes, invalid fixtures fail
- [x] Deploy with `ENABLED_PLUGINS=lvms,trust-manager` on a test cluster

## Summary by CodeRabbit

* **New Features**
  * Trust-manager plugin cluster issuer naming is now configurable. Users can customize the cluster issuer name per deployment instead of relying on hardcoded defaults, supporting diverse deployment scenarios.

* **Improvements**
  * Deployment procedures updated to support parameterized cluster issuer configurations, enhancing flexibility across different setup requirements.

### DIFF
--- a/plugins/trust-manager/defaults.yaml
+++ b/plugins/trust-manager/defaults.yaml
@@ -1,3 +1,4 @@
 ---
+trust_manager_cluster_issuer: default-ca
 trust_manager_ca_issuer_duration: 87600h
 trust_manager_ca_issuer_renew_before: 8760h
--- a/plugins/trust-manager/schemas/defaults.yaml
+++ b/plugins/trust-manager/schemas/defaults.yaml
@@ -3,9 +3,13 @@
 type: object
 additionalProperties: false
 required:
+  - trust_manager_cluster_issuer
   - trust_manager_ca_issuer_duration
   - trust_manager_ca_issuer_renew_before
 properties:
+  trust_manager_cluster_issuer:
+    "$ref": "#/definitions/nonEmptyString"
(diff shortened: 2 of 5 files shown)


maorfr (approved):

rporres:
what's the difference if we cannot override it in config? defaults are not supposed to be modified...

eurijon:
> what's the difference if we cannot override it in config? defaults are not supposed to be modified...

The PR title might be confusing as it says "externalize" when it's not externalizing it (which was my first idea). With this PR I'm creating the `trust_manager_cluster_issuer` variable in `defaults.yaml` as the variable is used in different places and also taken for certificate name.
At some point we might need to externalize it and make it configurable by the user but I'm avoiding to do it now until we see it's necessary.
I'm updating the PR title to better reflect the current changes.

rporres (approved):

eurijon:
The last push only amends the commit message, merging as it already passed all the checks before the commit amend

PR #440, opened 2026-06-03
fix cert-manager webhook OOM by increasing memory limit

eurijon:
## Summary

- The cert-manager webhook memory limit was set to 32Mi, causing OOM kills and admission webhook timeouts that broke CertificateRequest approval.
- Increased webhook memory from 32Mi to 128Mi (both request and limit).
- CPU unchanged (20m), Guaranteed QoS preserved (requests == limits).

## Test plan

- [x] Verified cert-manager webhook pods no longer OOM crash on openshift-qe-002
- [x] Verified CertificateRequest approval works correctly
- [x] Verified ClusterIssuer reaches Ready state

### DIFF
--- a/operators/openshift-cert-manager-operator/tasks.yaml
+++ b/operators/openshift-cert-manager-operator/tasks.yaml
@@ -38,11 +38,11 @@
               overrideReplicas: "{{ cert_manager_webhook_replicas }}"
               overrideResources:
                 limits:
-                  cpu: 500m
+                  cpu: 20m
                   memory: 128Mi
                 requests:
-                  cpu: 250m
-                  memory: 100Mi
+                  cpu: 20m
+                  memory: 128Mi
 
     - name: Read back CertManager CR
       kubernetes.core.k8s_info:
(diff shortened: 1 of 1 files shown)


maorfr (approved):
would it make sense to make this configurable for the consumer?

eurijon:
> would it make sense to make this configurable for the consumer?

We can but adding so many params can complicate initial enclave configuration. We have a lot of "low level" operator config parameters that are not exposed yet to user, what's your opinion?
I would start from some default values and see the behavior in our tests and production, and then make them configurable if needed.

operators/openshift-cert-manager-operator/tasks.yaml
```diff
@@ -38,11 +38,11 @@
               overrideReplicas: "{{ cert_manager_webhook_replicas }}"
               overrideResources:
                 limits:
-                  cpu: 20m
-                  memory: 32Mi
+                  cpu: 500m
```
rporres: why increasing cpu limits if we're having OOM, not CPU throttling (per PR description).

If we have to increase cpu limits, why not 250m to keep the `Guaranteed` QoS the previous settings implied?

eurijon: Good point, the cpu limits seemed to be low, but leaving untouched for now. Also memory limits and requests are set to 128Mi to keep guaranteed QoS - commit https://github.com/rh-ecosystem-edge/enclave/pull/440/changes/4e606f8cf30338badca479088b5ce26738f727b0

rporres (approved):

PR #435, opened 2026-06-03
OSAC-1240: Add declarative Kubernetes resource dependency validation to plugins

maorfr:
## Summary

Implements declarative Kubernetes resource dependency validation for plugins. Plugins can now declare required K8s resources in `plugin.yaml` instead of writing manual validation tasks.

## Changes

- **Schema**: Extended `schemas/plugin.yaml` with `requires.resources` field
- **Validation**: Added automated `k8s_info` validation in `playbooks/tasks/deploy_plugin.yaml`
- **Migrations**: 
  - `trust-manager`: Removed manual CRD/Deployment checks (27 lines deleted)
  - `example`: Extracted namespace check to declarative `requires.resources`
- **Documentation**: Added comprehensive design doc at `docs/design/plugin-resource-dependencies.md`

## Example Usage

```yaml
requires:
  resources:
    - apiVersion: apiextensions.k8s.io/v1
      kind: CustomResourceDefinition
      name: certificates.cert-manager.io
    - apiVersion: apps/v1
      kind: Deployment
      name: cert-manager
      namespace: cert-manager
    - apiVersion: v1
      kind: Namespace
      name: default
```

## Benefits

- **DRY**: Eliminates repetitive validation boilerplate from plugin tasks
- **Flexible**: Works for any Kubernetes resource (CRDs, Deployments, Namespaces, ConfigMaps, etc.)
- **Self-documenting**: Dependencies are explicit and visible in `plugin.yaml`
- **Consistent errors**: All validation failures use the same clear message format
- **Backward compatible**: Existing plugins work unchanged

## Testing

- ✅ Schema validation passed: `make -f Makefile.ci validate-plugins`
- ✅ Plugin validation passed for all 10 plugins
- ✅ trust-manager now uses declarative validation
- ✅ example plugin demonstrates namespace check pattern

## Files Changed

7 files changed, 411 insertions(+), 39 deletions(-)

## Tracking

Jira: [OSAC-1240](https://redhat.atlassian.net/browse/OSAC-1240)

## Summary by CodeRabbit

* **New Features**
  * Plugins can declaratively declare required Kubernetes resources; installer validates those resources in-cluster and fails with clear messages when missing.
  * Conditional resource checks supported via Jinja2-style `when`.

* **Documentation**
  * Added design and architecture docs with examples, validation behavior, migration guidance, and testing scenarios.

* **Refactor**
  * Removed duplicated manual pre-validation checks in favor of the centralized resource-validation step.

PR #443, opened 2026-06-04
Fix file validation when condition re-evaluation

maorfr:
## Summary

Fixes a pre-existing issue where the file validation task re-evaluates `item.when` instead of checking whether the stat task actually ran.

## Problem

The current code loops over `plugin.requires.files` and checks `item.when | default(true) | bool`, which re-evaluates the condition. This could lead to inconsistency if variables change between the stat task and the validation task.

## Solution

- Loop over `r_required_files.results` directly instead of using `ansible_loop.index0`
- Check `not item.skipped | default(false)` to rely on the actual stat task outcome
- Simplifies the code and makes it more reliable

## Changes

```yaml
# Before
loop: "{{ plugin.requires.files | default([]) }}"
when:
  - item.when | default(true) | bool
that: r_required_files.results[ansible_loop.index0].stat.exists

# After  
loop: "{{ r_required_files.results }}"
when:
  - not item.skipped | default(false)
that: item.stat.exists
```

This matches the pattern that will be used in #435 for resource validation.

## Testing

- [ ] Existing plugins with file requirements still validate correctly
- [ ] Conditional file requirements (with `when` clauses) work as expected

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/playbooks/tasks/deploy_plugin.yaml
+++ b/playbooks/tasks/deploy_plugin.yaml
@@ -107,19 +107,18 @@
 
 - name: Validate plugin requirements (files)
   ansible.builtin.assert:
-    that: r_required_files.results[ansible_loop.index0].stat.exists
+    that: item.stat.exists
     fail_msg: >-
-      Plugin '{{ plugin.name }}' requires file '{{ item.path }}' but it was not found at
-      {{ plugin_dir }}/{{ item.path }}.
-      {{ item.description | default('') }}
-  loop: "{{ plugin.requires.files | default([]) }}"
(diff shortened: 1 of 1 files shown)


eurijon (commented):
Ignore this comment (by claude)

eurijon (commented):
Should we apply the same loop pattern to https://github.com/rh-ecosystem-edge/enclave/blob/5610402a390d017ec47c54184ce7931a49c60a21/playbooks/tasks/validate_single_plugin.yaml#L56-L81?

PR #458, opened 2026-06-08
Add macAddress as mandatory to schema validation

javipolo:
macAddress should be mandatory because it is used by ironic

it also cannot be extracted consistently from mapInterfaces, so it should remain a mandatory field in each host entry


rporres:
Can you add a commit message explaining why this change is needed?

javipolo:
@rporres sure

I added a concise explanation in the commit message, and explained a bit more in the PR description!

rporres (approved):

eurijon (approved):

PR #459, opened 2026-06-09
move Mirrored catalogsource configuration ACM policy to operators

maorfr:
this step is currently executed as part of day 2 operations, but without it - core operators can not be installed on spoke clusters. moving it to be installed after operators.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized installation workflow to apply ACM catalog source policies during post-install phase rather than day-2 operations, simplifying the day-2 workflow to focus on Clair disconnected configuration.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -266,8 +266,7 @@ step_operators() {
 }
 
 step_day2() {
-    echo "Catalog source ACM policy .." | tee -a ${log}
-    ANSIBLE_LOG_PATH=${log} ansible-playbook playbooks/06-day2.yaml $EXTRA_VARS --tags acm-policy-catalogsources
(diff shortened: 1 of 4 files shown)


eurijon (commented):
ACM is installed in phase 5 (`playbooks/05-operators.yaml`) so shouldn't the mirrored catalogsource policy be placed after ACM is fully set up? before addon plugins or as part of operator tasks?

maorfr:
> ACM is installed in phase 5 (`playbooks/05-operators.yaml`) so shouldn't the mirrored catalogsource policy be placed after ACM is fully set up? before addon plugins or as part of operator tasks?

you are very right. implemented exactly as you described.

maorfr:
e2e failure related to #469

rporres (approved):

PR #467, opened 2026-06-10
Approve pending upgrade InstallPlans during MCH wait

eliorerz:
## Summary

- When MCE is installed with Manual approval, a newer version can appear in the catalog while MCH is reconciling. This creates a second InstallPlan in `RequiresApproval` phase that nobody approves, causing MCH to get stuck in `phase: Installing` forever.
- Replace the simple `k8s_info` MCH wait with a shell task that directly approves any unapproved InstallPlans in the MCE and ACM namespaces on every retry iteration before checking MCH phase. This bypasses version gating since the version-gated reconcile tool would skip InstallPlans for versions newer than pinned.

### DIFF
--- a/operators/advanced-cluster-management/tasks.yaml
+++ b/operators/advanced-cluster-management/tasks.yaml
@@ -13,22 +13,32 @@
   register: r_mhc
   until: r_mhc is succeeded
 
-- name: Wait until RHACM MultiClusterHub is ready
-  kubernetes.core.k8s_info:
-    api_version: operator.open-cluster-management.io/v1
-    kind: MultiClusterHub
-    name: multiclusterhub
-    namespace: open-cluster-management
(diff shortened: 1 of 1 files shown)


maorfr:
this PR effectively cancelles version pinning, not sure we can merge it?

maorfr:
✨ **Claude Code**:

I've created an alternative proposal in PR #471 that introduces a more flexible  property for operators.

Instead of hardcoding logic for multicluster-engine, the new approach:
- Adds a  property to the operator definition
- Uses automatic install plan approval when  and 
- Is extensible to any operator in the future

This is proposed as a workaround. We should still pursue a bug with the ACM team for a proper fix, but this provides a configurable solution in the meantime.

See: https://github.com/rh-ecosystem-edge/enclave/pull/471

PR #469, opened 2026-06-10
Update multicluster-engine version to 2.10.3

maorfr:
bump version to fix connected mode. similar to #331 

## Chores

* Updated multicluster-engine operator to version 2.10.3


maorfr:
related to #467

rporres (approved):

PR #471, opened 2026-06-10
Add connected_auto property for operators to enable automatic approval when not disconnected

maorfr:
## Summary
This change introduces a new `connected_auto` property for operators that enables automatic install plan approval when the environment is not disconnected.

## Changes
- Added `connected_auto: true` to multicluster-engine operator in `defaults/operators.yaml`
- Updated `schemas/operators.yaml` to include the new boolean property with validation
- Modified `playbooks/tasks/configure_operator.yaml` to:
  - Extract approval logic to `_use_automatic_approval` variable
  - Use Automatic approval when operator has `connected_auto: true` and not disconnected
  - Maintain Manual approval for all other cases

## How it works
- When `disconnected: false` and operator has `connected_auto: true`, the operator subscription uses `installPlanApproval: Automatic`
- Otherwise, operators in custom namespaces use `installPlanApproval: Manual` (existing behavior)
- Operators in `openshift-operators` namespace always use `Automatic` (existing behavior)

## Extensibility
This approach is extensible - any operator can set `connected_auto: true` in the future to get automatic approval when connected, while maintaining the current Manual behavior in disconnected environments.

## Note
This is a workaround proposal to replace PR #467. We should pursue a bug with the ACM team for a proper fix, but this provides a flexible solution in the meantime.

### DIFF
--- a/operators/advanced-cluster-management/configure_mch.yaml
+++ b/operators/advanced-cluster-management/configure_mch.yaml
@@ -13,11 +13,11 @@
 - name: Set MCE subscription spec for connected
   when: not (disconnected | default(true) | bool)
   vars:
-    _mce_operator: "{{ operators | selectattr('name', 'equalto', 'multicluster-engine') | first }}"
+    _mce_operator: "{{ (operators | selectattr('name', 'equalto', 'multicluster-engine') | list)[0] | default({}) }}"
     _mce_use_automatic: "{{ _mce_operator.connected_auto | default(false) | bool }}"
   ansible.builtin.set_fact:
     mce_subscription_spec:
-      installPlanApproval: "{{ 'Automatic' if _mce_use_automatic else 'Manual' }}"
+      installPlanApproval: "{{ 'Automatic' if (_mce_use_automatic | bool) else 'Manual' }}"
 
 - name: Ensure RHACM MultiClusterHub is present with MCE subscription spec
   kubernetes.core.k8s:
--- a/operators/advanced-cluster-management/configure_mch.yaml
+++ b/operators/advanced-cluster-management/configure_mch.yaml
@@ -12,9 +12,12 @@
 
 - name: Set MCE subscription spec for connected
   when: not (disconnected | default(true) | bool)
+  vars:
+    _mce_operator: "{{ operators | selectattr('name', 'equalto', 'multicluster-engine') | first }}"
+    _mce_use_automatic: "{{ _mce_operator.connected_auto | default(false) | bool }}"
   ansible.builtin.set_fact:
     mce_subscription_spec:
-      installPlanApproval: Manual
+      installPlanApproval: "{{ 'Automatic' if _mce_use_automatic else 'Manual' }}"
 
 - name: Ensure RHACM MultiClusterHub is present with MCE subscription spec
   kubernetes.core.k8s:
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -75,10 +75,14 @@
     namespace: "{{ operator.namespace }}"
   register: r_sub_check
 
+- name: "Determine if operator should use automatic approval (not in disconnected)"
+  ansible.builtin.set_fact:
+    _use_automatic_approval: "{{ operator.connected_auto | default(false) and not (disconnected | default(true) | bool) }}"
+
 - name: "Ensure Subscription exists (Automatic)"
   when:
     - r_sub_check.resources | length == 0
-    - operator.namespace == "openshift-operators"
+    - operator.namespace == "openshift-operators" or _use_automatic_approval
   kubernetes.core.k8s:
     state: present
     definition:
@@ -102,6 +106,7 @@
   when:
     - r_sub_check.resources | length == 0
     - operator.namespace != "openshift-operators"
+    - not _use_automatic_approval
   kubernetes.core.k8s:
     state: present
     definition:
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -75,14 +75,16 @@
     namespace: "{{ operator.namespace }}"
   register: r_sub_check
 
-- name: "Determine if operator should use automatic approval (not in disconnected)"
+- name: "Determine if operator should use automatic approval"
   ansible.builtin.set_fact:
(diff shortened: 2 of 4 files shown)


maorfr:
✨ **Claude Code**:

Updated the PR with an additional commit (aed3e9d) to handle the ACM MultiClusterHub configuration.

The `operators/advanced-cluster-management/configure_mch.yaml` file was hardcoding `installPlanApproval: Manual` for MCE even in connected mode. This is now updated to respect the `connected_auto` property, ensuring consistent behavior across both direct MCE installation and ACM-managed MCE installation.

**Changes in this update:**
- `operators/advanced-cluster-management/configure_mch.yaml`: Now uses Automatic approval when `connected_auto: true` and not disconnected

The branch has been rebased on main to pick up recent changes.

operators/advanced-cluster-management/configure_mch.yaml
```diff
@@ -12,9 +12,12 @@
 
 - name: Set MCE subscription spec for connected
   when: not (disconnected | default(true) | bool)
+  vars:
+    _mce_operator: "{{ operators | selectattr('name', 'equalto', 'multicluster-engine') | first }}"
+    _mce_use_automatic: "{{ _mce_operator.connected_auto | default(false) | bool }}"
   ansible.builtin.set_fact:
     mce_subscription_spec:
-      installPlanApproval: Manual
+      installPlanApproval: "{{ 'Automatic' if _mce_use_automatic else 'Manual' }}"
```
eurijon:
```suggestion
      installPlanApproval: "{{ 'Automatic' if (_mce_use_automatic | bool) else 'Manual' }}
```
Better to cast it to bool as "False" can be interpreted as true (non-empty string).

playbooks/tasks/configure_operator.yaml
```diff
@@ -75,10 +75,14 @@
...
 
+- name: "Determine if operator should use automatic approval (not in disconnected)"
+  ansible.builtin.set_fact:
+    _use_automatic_approval: "{{ operator.connected_auto | default(false) and not (disconnected | default(true) | bool) }}"
+
 - name: "Ensure Subscription exists (Automatic)"
   when:
     - r_sub_check.resources | length == 0
-    - operator.namespace == "openshift-operators"
+    - operator.namespace == "openshift-operators" or _use_automatic_approval
```
eurijon:
```suggestion
    - operator.namespace == "openshift-operators" or (_use_automatic_approval | bool)
```
same here

playbooks/tasks/configure_operator.yaml
```diff
@@ -102,6 +106,7 @@
   when:
     - r_sub_check.resources | length == 0
     - operator.namespace != "openshift-operators"
+    - not _use_automatic_approval
```
eurijon:
```suggestion
    - not (_use_automatic_approval | bool)
```
Same here

operators/advanced-cluster-management/configure_mch.yaml
```diff
@@ -12,9 +12,12 @@
 
 - name: Set MCE subscription spec for connected
   when: not (disconnected | default(true) | bool)
+  vars:
+    _mce_operator: "{{ operators | selectattr('name', 'equalto', 'multicluster-engine') | first }}"
```
rporres: Maybe (a bit defensive but just in case)

```
(operators | selectattr('name', 'equalto', 'multicluster-engine') | list)[0] | default({})
```

playbooks/tasks/configure_operator.yaml
rporres: We don't want to approve the plan  if `_use_automatic_approval`, right? We would need to add
```suggestion
 - not _use_automatic_approval
```

playbooks/tasks/configure_operator.yaml
```diff
@@ -75,10 +75,14 @@
     namespace: "{{ operator.namespace }}"
   register: r_sub_check
 
+- name: "Determine if operator should use automatic approval (not in disconnected)"
+  ansible.builtin.set_fact:
+    _use_automatic_approval: "{{ operator.connected_auto | default(false) and not (disconnected | default(true) | bool) }}"
```
rporres: I guess this is the place to fix @eurijon's concerns for `_use_automatic_approval`

```suggestion
    _use_automatic_approval: "{{ operator.connected_auto | default(false) | bool and not (disconnected | default(true) | bool) }}"
```

maorfr:
@coderabbitai review

operators/advanced-cluster-management/configure_mch.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c — added `| bool` cast to the ternary.

playbooks/tasks/configure_operator.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c. Also folded the `openshift-operators` namespace check into `_use_automatic_approval` itself, so this condition is now just `_use_automatic_approval | bool`.

playbooks/tasks/configure_operator.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c — now `not (_use_automatic_approval | bool)`.

operators/advanced-cluster-management/configure_mch.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c — using `(operators | selectattr('name', 'equalto', 'multicluster-engine') | list)[0] | default({})` as suggested.

playbooks/tasks/configure_operator.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c. Also extended the same guard to "Wait for Install Plan" and "Print Install Plan approval output", and widened "Get Installed CSV"/"Wait until CSV is installed" to run for auto-approved operators too, so they are still waited on instead of being skipped entirely.

playbooks/tasks/configure_operator.yaml
maorfr: ✨ **Claude Code**: Fixed in 283ce3c. Went one step further and folded the `openshift-operators` namespace check into `_use_automatic_approval` too, so it is now the single source of truth for every subscription/install-plan/CSV task in this file.

eurijon (approved):

rporres (approved):

PR #473, opened 2026-06-10
Move Python source into enclave namespace package

rporres:
All modules (cli, utils, reconcile/, tools/) are relocated from the flat src/ root into src/enclave/, so the installed wheel places them under site-packages/enclave/ instead of directly in site-packages.

This eliminates name collision risk with generic top-level names like 'utils', 'reconcile', and 'tools'. Tests are excluded from the wheel by narrowing only-include to src/enclave/.

OSAC-1396

## Summary by CodeRabbit

* **Chores**
  * Reorganized internal package structure with updated imports throughout codebase
  * Fixed default configuration file resolution for consistent behavior across installation types
  * Updated CLI entry point and packaging configuration to reflect restructured organization


maorfr (approved):

rporres:
error in https://github.com/rh-ecosystem-edge/enclave/actions/runs/27293821310/job/80621204897?pr=473 is unrelated, merging.

PR #474, opened 2026-06-11
test openshift-ai plugin installation

maorfr:
## Summary by CodeRabbit

* **New Features**
  * OpenShift AI plugin is now enabled by default in the deployment configuration.

PR #357, opened 2026-05-14
Auto-deploy addon plugins during deploy-cluster

eliorerz:
## Summary
- Remove `type: foundation` filter from `collect_core_plugin_operators.yaml` and `collect_plugin_registries.yaml` so all enabled plugins (foundation + addon) are mirrored in a single oc-mirror pass
- Add `deploy_addon_plugins.yaml` to auto-deploy addon plugins (e.g. nvidia-gpu, openshift-ai) after core operators and Quay mirroring complete
- Add addon deployment step to `05-operators.yaml` at the end of Phase 5

## Release Notes

* **New Features**
  * Added automatic discovery and deployment of addon plugins in Phase 5, occurring after core operators and Quay mirroring completion.
  * Registry mirroring configuration now includes all enabled plugins (foundation and addon).

[Review Change Stack](https://app.coderabbit.ai/change-stack/rh-ecosystem-edge/enclave/pull/357)

PR #476, opened 2026-06-11
do not auto-deploy addon plugins

maorfr:
introduced in #357, this functionality is not performing as expected.

this was tested in #474: https://github.com/rh-ecosystem-edge/enclave/actions/runs/27323271783/job/80718919973?pr=474

this is a temporary removal as we want to support this feature.

### DIFF
--- a/playbooks/05-operators.yaml
+++ b/playbooks/05-operators.yaml
@@ -85,17 +85,3 @@
             apply:
               tags: operators
           tags: operators
-
-        - name: Auto-discover and deploy addon plugins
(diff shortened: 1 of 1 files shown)


maorfr:
related to:
- https://github.com/gori-project/GoRI/issues/903
- https://github.com/gori-project/GoRI/issues/911

rporres:
While it is broken, I would add a validation to prevent anyone from attempting this. It is not obvious this is going to break and somebody may want to install an addon plugin at bootstrap

PR #483, opened 2026-06-14
Add validation to detect persistent Metal3 from previous deployments

maorfr:
## Summary

Adds early validation to prevent confusing 401 Unauthorized errors when persistent Metal3 components from a previous deployment are still running.

## Problem

The real root cause of GoRI issue #915 was **not** the htpasswd generation, but leftover Metal3 infrastructure from a previous deployment:

1. Previous deployment used `metal3_persistent: true` → creates systemd quadlet services running as **root**
2. Cleanup script only removed **user-level** podman resources
3. New deployment generates new credentials in user secret store
4. New user-level Ironic container fails to start (port already bound by root Ironic)
5. API calls hit the **old root-level Ironic** with old credentials → 401 Unauthorized

This caused hours of debugging because:
- Password verification showed credentials were correct ✓
- Container saw correct hash ✓
- But authentication failed ✗ (hitting wrong Ironic instance)

## Solution

Added `validate_no_persistent_metal3.yaml` that runs before Metal3 deployment and checks for:
- Root-level Metal3 pods (`sudo podman pod ps`)
- Metal3 systemd services (quadlets)
- Quadlet unit files in `/etc/containers/systemd/`

**Fails fast** with clear error message including:
- Explanation of the problem
- List of what was detected
- Complete cleanup commands (from Nick's solution)
- Reference to the GitHub issue

## Changes

1. **New file**: `playbooks/tasks/validate_no_persistent_metal3.yaml`
   - Validation task that checks for persistent Metal3
   - Comprehensive error message with cleanup instructions

2. **Modified**: `playbooks/tasks/configure_hardware_ironic_setup.yaml`
   - Added validation as first step before deploying Metal3

## Testing

Validation will detect the conditions that caused the issue on Cluster 10 and Cluster 17, failing early with actionable instructions instead of later with confusing 401 errors.

## Related Issues

- GoRI: [#915](https://github.com/gori-project/GoRI/issues/915)
- Root cause identified by @carbonin in [this comment](https://github.com/gori-project/GoRI/issues/915#issuecomment-4695481122)

## Checklist

- [x] Validation runs before Metal3 deployment
- [x] Error message includes complete cleanup commands
- [x] References the GitHub issue for context
- [x] Commit includes Assisted-by trailer

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment reliability by adding validation checks to detect and prevent conflicts from previous Metal3 installations. When legacy artifacts are detected, the deployment process now fails with clear error messages and specific cleanup instructions, ensuring successful redeployment.

### DIFF
--- a/playbooks/tasks/configure_hardware_ironic_setup.yaml
+++ b/playbooks/tasks/configure_hardware_ironic_setup.yaml
@@ -1,6 +1,9 @@
 ---
 # Set up Ironic for hardware configuration
 
+- name: Validate no persistent Metal3 from previous deployment
+  ansible.builtin.include_tasks: validate_no_persistent_metal3.yaml
+
 - name: Deploy Metal3 common resources
   ansible.builtin.include_tasks: deploy_metal3_common.yaml
 
--- a/playbooks/tasks/validate_no_persistent_metal3.yaml
+++ b/playbooks/tasks/validate_no_persistent_metal3.yaml
@@ -1,0 +1,89 @@
+---
+# Validate that no persistent Metal3 stack is running from a previous deployment
+# This prevents confusing 401 errors when old root-level Ironic containers
+# conflict with new user-level deployments
+
+- name: Check for running root-level Metal3 pod
+  become: true
+  ansible.builtin.command: podman pod ps --format {% raw %}'{{.Name}}'{% endraw %} --filter name=metal3-ironic
+  register: root_metal3_pods
+  changed_when: false
+  failed_when: false
+
+- name: Check for Metal3 systemd services
+  become: true
+  ansible.builtin.systemd:
+    name: "{{ item }}"
+  register: metal3_services
+  loop:
+    - metal3-ironic-pod.service
+    - metal3-ironic-api.service
+    - metal3-bmo.service
+    - metal3-httpd.service
+  failed_when: false
+  changed_when: false
+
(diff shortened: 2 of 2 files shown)


rporres:
I would make this more general and include previous registry installations in the LZ which also need to be cleaned-up. Something like using what we already have in RHDP env, the cleanup script.

FWIW, this alone is not going to solve the current issues there as I suspect the metal3 containers are restored by a different method there.

carbonin:
Instead of a validation could we implement some kind of forced clean slate variable that would actually do the wipe before we run?

On the same topic what are we expecting if we rerun bootstrap generally? Will the cluster redeploy or do we check if it's installed already? Do we fully re-mirror or do we check what's already present (maybe this is just a feature of `oc mirror`)?

I ask because we should be consistent. If we're expecting a rerun of bootstrap to redeploy everything then we should be cleaning out these pods ourselves unconditionally. If we want a rerun to do more of a "desired state" thing, then we definitely _shouldn't_ clean anything out if it already exists.

Does that make sense?

rporres:
closing this, it will be taken care differently in #496 (still a WIP, though)

rporres:
@carbonin The idea is to check if a wipe is needed in bootstrap, using the same philosophy as the rest of the validations, asking the user to perform the change before actually doing it.

PR #484, opened 2026-06-14
Load plugin user config for day-2 deployments

maorfr:
## Summary

- Add loading of plugin-specific user configuration files (`config/plugins/<name>.yaml`) for day-2 plugin deployments via `deploy-plugin.yaml`
- The existing `load-vars.yaml` task only loads config files for plugins in `enabled_plugins`, but day-2 plugins deployed via `deploy-plugin.yaml` may not be in that list
- This ensures plugin user config is available regardless of deployment method

## Test plan

- [ ] Day-2 plugin deployment with user config file loads successfully
- [ ] Day-2 plugin deployment without user config file continues to work
- [ ] `make -f Makefile.ci validate` passes

### DIFF
--- a/playbooks/tasks/deploy_plugin.yaml
+++ b/playbooks/tasks/deploy_plugin.yaml
@@ -78,6 +78,21 @@
     - not r_plugin_defaults_yaml.stat.exists
   tags: always
 
+# Load user config for this plugin (config/plugins/<name>.yaml) if it exists.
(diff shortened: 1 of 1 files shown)


maorfr:
picked from https://github.com/rh-ecosystem-edge/enclave/pull/482/changes#diff-6f9767dfd8e0bfd6da8ec3238f1c1b382ecd2863fd563842cb38631ad966250e

PR #490, opened 2026-06-15
Make quayPinnedImageSetEnabled opt-in instead of opt-out

maorfr:
## Summary
Changes the `quayPinnedImageSetEnabled` feature from opt-out (enabled by default) to opt-in (disabled by default).

## Context
Per Slack discussion, this feature should be opt-in to avoid blocking users who encounter issues with it. Users can explicitly enable it when needed.

## Changes
- Set default to `false` in `schemas/global.yaml`
- Simplified conditional logic in `operators/quay-operator/quay_disconnected_finish.yaml`
- Removed backward compatibility variable `quay_pinnedimageset_enabled`
- Updated documentation in `CONFIGURATION_REFERENCE.md` to reflect opt-in behavior

## Testing
- Verified schema default is now `false`
- Verified playbook conditional properly uses the new default
- Documentation accurately describes the opt-in behavior

## Related Issues
- Addresses feedback from GoRI issue #921

### DIFF
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -322,23 +322,6 @@ defaultNtpServers:
 - Only needed when cluster nodes cannot reach the default public NTP pool
 - Useful in air-gapped or firewalled environments
 
-#### `quayPinnedImageSetEnabled`
-
-**Description**: Controls whether the post-mirror master `PinnedImageSet` is generated and applied after Quay image mirroring.
-
-**Type**: Boolean (optional)
-
-**Default**: `true`
-
-**Example**:
-```yaml
-quayPinnedImageSetEnabled: false
-```
-
-**Notes**:
-- Set to `false` when you want to skip post-mirror prefetch pinning.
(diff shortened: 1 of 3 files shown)


maorfr:
## Relation to GoRI Issue #921

This PR fixes https://github.com/gori-project/GoRI/issues/921

GoRI #921 reports a bootstrap failure where master node image digest collection failed for one node (`control2`), causing the entire deployment to fail with:
```
Master node image digest collection failed or produced no usable digest refs.
Failed nodes: [] Nodes with empty output: ['control2']
```

**Root cause**: The error originates from the PinnedImageSet generation process in `collect_master_node_image_digests.yaml`, which was **enabled by default** (opt-out). When the `enclave tools collect-node-image-digests` command failed or returned no output for a single node, it blocked the entire bootstrap.

**Fix**: By making `quayPinnedImageSetEnabled` opt-in (default: `false`), deployments will no longer be blocked by transient node collection issues unless users explicitly enable the feature. This matches the workaround suggested in the Slack thread (manually setting `quayPinnedImageSetEnabled: false`), but now it's the default behavior.

PR #491, opened 2026-06-15
Use default_operator_source to configure MultiClusterHub

maorfr:
## Summary
Replace hardcoded catalog source name with the `default_operator_source` variable in MCE subscription configuration.

## Context
Follow-up to #488 (Extract MultiClusterHub configuration to reusable task).

## Changes
- Changed MCE subscription source from hardcoded `cs-mirror-redhat-operators-v4-20` to `{{ default_operator_source }}`

## Benefits
- Consistent with other operator configurations that use `default_operator_source`
- Avoids hardcoding version-specific catalog source names (the `-v4-20` suffix)
- Makes the configuration more flexible across different OpenShift versions

## Testing
- Verified the variable reference is correct
- Consistent with existing operator configuration patterns

## Release Notes

* **Bug Fixes**
  * Improved operator source configuration for disconnected environments by enabling dynamic defaults instead of hardcoded values, providing greater deployment flexibility.

### DIFF
--- a/operators/advanced-cluster-management/configure_mch.yaml
+++ b/operators/advanced-cluster-management/configure_mch.yaml
@@ -8,7 +8,7 @@
   ansible.builtin.set_fact:
     mce_subscription_spec:
       installPlanApproval: Manual
-      source: cs-mirror-redhat-operators-v4-20
+      source: "{{ default_operator_source }}"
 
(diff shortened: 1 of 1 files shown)


maorfr:
this will fail when being called from sync.

PR #495, opened 2026-06-15
Fix foundation plugin catalog availability on spoke clusters

maorfr:
## Summary

Fixes foundation plugin (ODF, LVMS) catalog source availability on spoke clusters in disconnected environments. This is a regression from the May 2026 catalog separation where foundation plugins moved to dedicated catalogs but these were not being deployed to spoke clusters.

**Fixes**: https://github.com/gori-project/GoRI/issues/922 (Sev 1 stop-ship bug)

## Problem

After the May 2026 catalog separation (commits 761e628, b44e490), foundation plugins got plugin-specific catalog sources (e.g., `cs-mirror-redhat-operators-odf-v4-20`) instead of being part of the core catalog. However, these plugin-specific catalogs were not being deployed to spoke clusters via ACM policies, breaking operator availability.

**Before**: Foundation plugins mirrored in core catalog → core catalog pushed to all clusters → operators available everywhere

**After catalog separation**: Foundation plugins in dedicated catalogs → only deployed to management cluster → operators unavailable on spoke clusters

## Solution

1. **Add `clusterSelector`** to ODF and LVMS plugins to target all managed clusters (excluding management cluster)

2. **Introduce `installOperatorsFleet` field** to control whether operator subscriptions are deployed to fleet via ACM policies
   - Defaults to `true` for backward compatibility
   - Set to `false` for ODF/LVMS (catalog-only deployment)

3. **Enhance fleet deployment logic** to conditionally create operator policies based on `installOperatorsFleet`

4. **Add schema dependency**: `installOperatorsFleet` requires `clusterSelector`

5. **Add migration** for existing deployments (runs during `upgrade.sh`)

### Plugin Configuration

**`plugins/odf/plugin.yaml`** and **`plugins/lvms/plugin.yaml`**:
```yaml
clusterSelector:
  matchExpressions:
    - key: local-cluster
      operator: NotIn
      values:
        - "true"

installOperators: true           # Install on management cluster
installOperatorsFleet: false     # Catalog only on spoke clusters
```

### Core Infrastructure

- **`playbooks/tasks/configure_operators_fleet.yaml`**: Conditional operator policy creation
- **`files/plugin-fleet-configuration/30-placementbinding.yaml.j2`**: Conditional subject binding (catalogsource only vs catalogsource + operators)
- **`schemas/plugin.yaml`**: Add `installOperatorsFleet` field with `clusterSelector` dependency
- **`scripts/verification/validate_plugins.sh`**: Add to valid fields

### Migration

- **`playbooks/tasks/migrations/foundation_plugin_fleet_catalogs.yaml`**: Creates ACM policies for existing deployments
- **`playbooks/tasks/migrations.yaml`**: Register the migration

## Migration Note for Existing Deployments

This change restores catalog source availability on spoke clusters but does **NOT** migrate existing operator Subscriptions on spoke clusters. If you have ODF/LVMS operators already installed on spoke clusters with Subscriptions pointing to the old core catalog (`cs-mirror-redhat-operators-v4-20`), you will need to migrate those Subscriptions to the new plugin-specific catalog sources (`cs-mirror-redhat-operators-odf-v4-20` or `cs-mirror-redhat-operators-lvms-v4-20`). Enclave consumers should handle this migration in their own upgrade processes. See `playbooks/tasks/migrations/foundation_plugin_catalog_sources.yaml` for the migration pattern used on the management cluster.

## Testing

- ✅ Schema validation passes for all plugins
- ✅ ODF plugin validates with new fields
- ✅ LVMS plugin validates with new fields
- ✅ Migration is self-contained and reuses existing infrastructure

## Result

ODF and LVMS catalog sources are now available on spoke clusters in disconnected environments, restoring pre-separation behavior while maintaining the benefits of per-plugin catalog isolation.

### DIFF
--- a/playbooks/tasks/configure_operators_fleet.yaml
+++ b/playbooks/tasks/configure_operators_fleet.yaml
@@ -53,7 +53,7 @@
   kubernetes.core.k8s:
     state: present
     namespace: openshift-acm-policies
-    definition: "{{ lookup('ansible.builtin.template', 'files/plugin-fleet-configuration/30-placementbinding.yaml.j2') | from_yaml }}"
+    definition: "{{ lookup('ansible.builtin.template', '{{ playbook_dir }}/../files/plugin-fleet-configuration/30-placementbinding.yaml.j2') | from_yaml }}"
   register: r_acm_placementbinding
   retries: "{{ k8s_retries }}"
   delay: "{{ k8s_delay }}"
--- a/playbooks/tasks/configure_operators_fleet.yaml
+++ b/playbooks/tasks/configure_operators_fleet.yaml
@@ -32,6 +32,8 @@
   retries: "{{ k8s_retries }}"
   delay: "{{ k8s_delay }}"
   until: r_acm_policy_operators is success
+  when:
+    - plugin.installOperatorsFleet | default(true) | bool
 
 - name: Create ACM PlacementRule for Plugin
   environment:
--- a/playbooks/tasks/configure_operators_fleet.yaml
+++ b/playbooks/tasks/configure_operators_fleet.yaml
@@ -3,7 +3,7 @@
     KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
   kubernetes.core.k8s:
     state: present
-    definition: "{{ lookup('ansible.builtin.file', 'files/plugin-fleet-configuration/00-namespace.yaml') | from_yaml }}"
+    definition: "{{ lookup('ansible.builtin.file', '{{ playbook_dir }}/../files/plugin-fleet-configuration/00-namespace.yaml') | from_yaml }}"
   register: r_acm_policy_ns
   retries: "{{ k8s_retries }}"
   delay: "{{ k8s_delay }}"
@@ -15,7 +15,7 @@
   kubernetes.core.k8s:
     state: present
     namespace: openshift-acm-policies
-    definition: "{{ lookup('ansible.builtin.template', 'files/plugin-fleet-configuration/09-policy-catalogsource.yaml.j2') | from_yaml }}"
+    definition: "{{ lookup('ansible.builtin.template', '{{ playbook_dir }}/../files/plugin-fleet-configuration/09-policy-catalogsource.yaml.j2') | from_yaml }}"
   register: r_acm_policy_cs
   retries: "{{ k8s_retries }}"
   delay: "{{ k8s_delay }}"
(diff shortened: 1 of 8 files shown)


javipolo (approved):

playbooks/tasks/migrations/foundation_plugin_fleet_catalogs.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+- name: Create ACM policies for foundation plugin fleet catalogs
+  ansible.builtin.include_tasks:
+    file: tasks/configure_operators_fleet.yaml
+    apply:
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+  vars:
+    plugin: "{{ fleet_plugin }}"
+    catalog_image: "{{ (fleet_plugin.catalog | default('redhat')) == 'certified' | ternary(certified_operator_catalog, rh_operator_catalog) }}"
```
rporres, line 51: hmm, maybe this is better to honor the precedence rules in jinja

```suggestion
    catalog_image: "{{ ((fleet_plugin.catalog | default('redhat')) == 'certified') | ternary(certified_operator_catalog, rh_operator_catalog) }}"
```

and below lines too

playbooks/tasks/configure_operators_fleet.yaml
```diff
@@ -3,7 +3,7 @@
     KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
   kubernetes.core.k8s:
     state: present
-    definition: "{{ lookup('ansible.builtin.file', 'files/plugin-fleet-configuration/00-namespace.yaml') | from_yaml }}"
+    definition: "{{ lookup('ansible.builtin.file', '{{ playbook_dir }}/../files/plugin-fleet-configuration/00-namespace.yaml') | from_yaml }}"
```
rporres, line 6: I think it's better to avoid nested jinja and use `~` to concatenate
```suggestion
     definition: "{{ lookup('ansible.builtin.template', playbook_dir ~
  '/../files/plugin-fleet-configuration/09-policy-catalogsource.yaml.j2') | from_yaml }}"
```

schemas/plugin.yaml
```diff
@@ -238,6 +238,13 @@ properties:
   installOperators:
     type: boolean
     decsription: Install operators.
```
rporres, line 240: nit: Maybe worth to add here now that this only affects mgmt cluster (and fix the typo)

rporres (approved):
looks fine, a few nits here and there

eurijon (approved):
lgtm, just a minor note about adding the clusterSelector in the `plugin.yaml` file:
```
  clusterSelector:
    matchExpressions:
      - key: local-cluster
        operator: NotIn
        values:
          - "true"
```
looks like "all spokes except the management cluster" is the default so hardcoding it in each plugin might be redundant

maorfr:
> looks like "all spokes except the management cluster" is the default so hardcoding it in each plugin might be redundant

the default is an undefined clusterSelector which results in this functionality not being included. is that what you meant @eurijon?

eurijon:
> > looks like "all spokes except the management cluster" is the default so hardcoding it in each plugin might be redundant
> 
> the default is an undefined clusterSelector which results in this functionality not being included. is that what you meant @eurijon?

I meant the value of `clusterSelector`. Every plugin that enables fleet catalogs will likely use the same `NotIn local-cluster` expression. That common-case selector could be a default, so plugins only need to define `clusterSelector` when overriding it.

PR #500, opened 2026-06-15
OSAC-1427: Add Netris config passthrough to OSAC plugin

eliorerz:
## Summary

- Add `clusterFulfillment` and `networkFulfillment` sections to the OSAC Helm values template, conditionally rendered when `osacNetrisEnabled: true`
- Add 26 Netris-related fields to the plugin config schema with conditional `if/then` validation requiring core fields when enabled
- Add commented-out example config and `osacNetrisEnabled: false` default

## Test plan

- [x] `make -f Makefile.ci validate-plugins` passes
- [x] `make -f Makefile.ci validate-yaml` passes
- [x] `make -f Makefile.ci validate-ansible` passes
- [ ] Manual: render values template with Netris vars set, verify fulfillment sections appear
- [ ] Manual: render without `osacNetrisEnabled`, verify fulfillment sections are absent

## Summary by CodeRabbit

* **New Features**
  * Added optional NETRIS network/backend support to the OSAC plugin, including conditional Helm values for controller identity, steps collection, networking, routing, DNS/external access, and SSH/bastion connectivity.
  * Added a commented NETRIS configuration template example and introduced a new enablement flag in OSAC defaults.
* **Bug Fixes**
  * Extended OSAC configuration validation to require NETRIS controller/auth and tenant/site details when NETRIS is enabled, including mutual checks for AWS key/secret pairs.
* **Chores**
  * NETRIS remains disabled by default unless explicitly enabled.

### DIFF
--- a/plugins/osac/defaults.yaml
+++ b/plugins/osac/defaults.yaml
@@ -1,12 +1,15 @@
 ---
 # OSAC plugin defaults
-# Note: config/plugins/osac.yaml uses camelCase (osacProfilesList, osacBYODatabase)
+# Note: config/plugins/osac.yaml uses camelCase (osacProfilesList, osacBYODatabase, osacNetrisEnabled)
 # which is loaded into Ansible scope as-is. The defaults below use the same
 # camelCase convention to match. Ansible variables from defaults.yaml and
 # config/plugins/osac.yaml are merged into the same scope.
 osacProfilesList:
   - caas
   - vmaas
+
+# Netris network backend
+osacNetrisEnabled: false
 
 # AAP settings (chart manages the instance, these configure it)
 osac_aap_name: osac-aap
--- a/plugins/osac/schemas/config.yaml
+++ b/plugins/osac/schemas/config.yaml
@@ -45,5 +45,104 @@ properties:
     description: >-
       Optional container image overrides. When unset, the Helm chart's
       built-in image references are used.
+  osacNetrisEnabled:
+    type: boolean
+    description: "Enable Netris network backend for cluster/network fulfillment"
+  osacNetrisControllerUrl:
+    type: string
+    description: "Netris controller API URL (e.g. https://ctl.netris.io)"
+  osacNetrisUsername:
+    type: string
+    description: "Netris API username"
+  osacNetrisPassword:
+    type: string
+    description: "Netris API password"
+  osacNetrisSiteId:
+    type: string
+    description: "Netris site ID (integer as string)"
+  osacNetrisTenantId:
+    type: string
+    description: "Netris tenant ID (integer as string)"
+  osacNetrisTenantName:
+    type: string
+    description: "Netris tenant name"
+  osacNetrisStepsCollection:
+    type: string
+    description: "Ansible steps collection for Netris (default: netris.steps)"
+  osacNetrisMgmtVpcId:
+    type: string
+    description: "Management VPC ID in Netris"
+  osacNetrisMgmtVpcName:
+    type: string
+    description: "Management VPC name in Netris"
+  osacNetrisResourceClassMap:
+    type: string
+    description: "JSON dict mapping resource classes to Netris server cluster templates"
+  osacNetrisSshBastionHost:
+    type: string
+    description: "SSH bastion hostname/IP for server management access"
+  osacNetrisSshBastionUser:
+    type: string
+    description: "SSH bastion username"
+  osacNetrisSshUser:
+    type: string
+    description: "SSH username for bare-metal servers (typically 'core')"
+  osacNetrisMgmtRouteDestination:
+    type: string
+    description: "Management route CIDR (e.g. 10.8.0.0/30)"
+  osacNetrisMgmtRouteGateway:
+    type: string
+    description: "Management route gateway IP"
+  osacNetrisExternalAccessBaseDomain:
+    type: string
+    description: "DNS domain for cluster endpoints"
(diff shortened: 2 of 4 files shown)


plugins/osac/schemas/config.yaml
```diff
@@ -21,5 +21,97 @@ properties:
...
+    osacNetrisEnabled:
+      const: true
+then:
+  required:
+    - osacNetrisControllerUrl
+    - osacNetrisUsername
+    - osacNetrisPassword
+    - osacNetrisSiteId
+    - osacNetrisTenantId
+    - osacNetrisTenantName
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Confirmed — this is a real bug.

In JSON Schema, `if: { properties: { osacNetrisEnabled: { const: true } } }` evaluates to **true** when the property is absent, because `properties` only constrains properties that are *present*. An absent `osacNetrisEnabled` vacuously satisfies the constraint, so the `then: required: [...]` block fires for any config that omits the flag — incorrectly requiring all Netris fields even when Netris is not being configured.

Fix: add `required: [osacNetrisEnabled]` inside the `if` block so the condition only matches when the flag is both present and `true`:

```yaml
if:
  required:
    - osacNetrisEnabled
  properties:
    osacNetrisEnabled:
      const: true
then:
  required:
    - osacNetrisControllerUrl
    - osacNetrisUsername
    - osacNetrisPassword
    - osacNetrisSiteId
    - osacNetrisTenantId
    - osacNetrisTenantName
```

This is a blocker — existing configs that do not set `osacNetrisEnabled` will fail schema validation until it is fixed.

plugins/osac/templates/values.yaml.j2
```diff
@@ -100,3 +100,83 @@ dbMigrate:
...
+networkFulfillment:
+  enabled: true
+  config:
+    NETRIS_CONTROLLER_URL: "{{ osacNetrisControllerUrl }}"
+    NETRIS_USERNAME: "{{ osacNetrisUsername }}"
+    NETRIS_SITE_ID: "{{ osacNetrisSiteId }}"
+    NETRIS_TENANT_ID: "{{ osacNetrisTenantId }}"
+    NETRIS_TENANT_NAME: "{{ osacNetrisTenantName }}"
+  secret:
+    NETRIS_PASSWORD: "{{ osacNetrisPassword }}"
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: The duplication is intentional — `clusterFulfillment` and `networkFulfillment` are independent config blocks consumed by different services in the Helm chart, and the asymmetry between them (cluster gets many more optional SSH/VPC/route fields; network only needs the core connection params) makes a shared macro awkward. A `{% set %}` block would only work for individual variable values, not for the key-value pairs themselves, so it would not meaningfully reduce the duplication. Proposing to close this comment as by-design.

rporres (approved):
LGTM. Please address or close all coderabbitai comments before merging

plugins/osac/schemas/config.yaml
```diff
@@ -21,5 +21,104 @@ properties:
...
+    description: "AWS secret key for Route53 DNS management"
+  osacNetrisSshKey:
+    type: string
+    description: "SSH private key for bare-metal server access"
+  osacNetrisSshBastionKey:
+    type: string
+    description: "SSH private key for bastion host access"
 required:
   - osacAapLicenseFile
+dependencies:
```
eurijon, line 130: It looks like `dependencies`  was deprecated in favor of `dependentRequired`

plugins/osac/defaults.yaml
```diff
@@ -1,12 +1,13 @@
...
 # OSAC plugin defaults
-# Note: config/plugins/osac.yaml uses camelCase (osacProfilesList, osacBYODatabase)
+# Note: config/plugins/osac.yaml uses camelCase (osacProfilesList, osacBYODatabase, osacNetrisEnabled)
 # which is loaded into Ansible scope as-is. The defaults below use the same
 # camelCase convention to match. Ansible variables from defaults.yaml and
 # config/plugins/osac.yaml are merged into the same scope.
 osacProfilesList:
   - caas
   - vmaas
+osacNetrisEnabled: false
```
eurijon, line 12: Just cosmetic, can you add a separated section for Netris to keep other config format? i.e:
```
# Netris
osacNetrisEnabled: false
```

eurijon (commented):
It needs rebasing as there were some changes conflicting with the PR. The OSAC CI job should be passing now.

plugins/osac/defaults.yaml
eliorerz: ✨ **Claude Code**: Done in 168228a — added `# Netris network backend` section header with a blank line separator.

plugins/osac/schemas/config.yaml
eliorerz: ✨ **Claude Code**: `dependentRequired` is a JSON Schema 2019-09+ keyword. This schema explicitly declares `"$schema": "http://json-schema.org/draft-07/schema"` (added in main recently), and the project validates via `ansible.utils.jsonschema` which uses Draft 7. I verified that under Draft 7, `dependentRequired` is silently ignored — it produces zero validation errors — while `dependencies` is correctly enforced.

Keeping `dependencies` here to match the draft and ensure validation actually works. If the project moves to Draft 2019-09 in the future, this should be updated to `dependentRequired` at that time.

eurijon (approved):

PR #516, opened 2026-06-17
OSAC-1597: Use JSON patch to configure OSAC operator with AAP URL and token

eurijon:
## Summary

- `kubernetes.core.k8s` with `state: present` fails when patching the osac-operator deployment because the container spec omits the required `image` field
- Strategic merge patch (`--type=strategic`) also fails with the same validation error
- Replaced with `kubernetes.core.k8s_json_patch` using RFC 6902 `add` operations to append env vars without touching other container fields
- Also renamed the task to reflect both AAP URL and token being configured

## Test plan

- [x] Reproduced the failure with `oc apply` (missing image error)
- [x] Confirmed `oc patch --type=json` works
- [x] Deployed OSAC plugin end-to-end with this fix — operator patched and rolled out successfully
- [x] All post-validate checks passed (PostgreSQL, fulfillment services, operator, AAP, Authorino)

Fixes: [OSAC-1597](https://redhat.atlassian.net/browse/OSAC-1597)

## Summary by CodeRabbit

* **Chores**
  * Updated the OSAC post-Helm deployment workflow to patch the OSAC operator Deployment in-place (instead of replacing the full Deployment specification).
  * The operator now receives OSAC AAP connection settings via environment variables, including a new OSAC AAP token sourced from the cluster secret.
  * OSAC deploy completion output was adjusted to reflect the new patch-and-rollout behavior.

### DIFF
--- a/plugins/osac/tasks/deploy.yaml
+++ b/plugins/osac/tasks/deploy.yaml
@@ -128,49 +128,26 @@
         token: "{{ _osac_aap_token }}"
   no_log: true
 
-- name: Get current OSAC operator deployment
-  kubernetes.core.k8s_info:
-    api_version: apps/v1
-    kind: Deployment
-    name: osac-operator
-    namespace: osac
-  register: __r_osac_operator_deployment
-
-- name: Assert osac-operator has a single container
-  ansible.builtin.assert:
-    that:
-      - __r_osac_operator_deployment.resources[0].spec.template.spec.containers | length == 1
-    fail_msg: "osac-operator deployment has multiple containers — review env patch target"
-
-- name: Set OSAC operator env var names
-  ansible.builtin.set_fact:
-    _osac_operator_env_names: >-
-      {{ __r_osac_operator_deployment.resources[0].spec.template.spec.containers[0].env
-         | default([]) | map(attribute='name') | list }}
-
 - name: Patch OSAC operator deployment with AAP URL and AAP token
-  kubernetes.core.k8s_json_patch:
-    api_version: apps/v1
+  kubernetes.core.k8s:
+    state: patched
     kind: Deployment
     name: osac-operator
     namespace: osac
-    patch:
-      - op: add
-        path: /spec/template/spec/containers/0/env/-
-        value:
-          name: OSAC_AAP_URL
-          value: "{{ _osac_aap_url }}"
-      - op: add
-        path: /spec/template/spec/containers/0/env/-
-        value:
-          name: OSAC_AAP_TOKEN
-          valueFrom:
-            secretKeyRef:
-              name: osac-aap-secret
-              key: token
-  when:
-    - "'OSAC_AAP_URL' not in _osac_operator_env_names"
-    - "'OSAC_AAP_TOKEN' not in _osac_operator_env_names"
+    definition:
+      spec:
+        template:
+          spec:
+            containers:
+              - name: manager
+                env:
+                  - name: OSAC_AAP_URL
+                    value: "{{ _osac_aap_url }}"
+                  - name: OSAC_AAP_TOKEN
+                    valueFrom:
+                      secretKeyRef:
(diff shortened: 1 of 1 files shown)


plugins/osac/tasks/deploy.yaml
```diff
@@ -128,28 +128,26 @@
...
+    name: osac-operator
+    namespace: osac
+    patch:
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
+        value:
+          name: OSAC_AAP_URL
+          value: "{{ _osac_aap_url }}"
+      - op: add
+        path: /spec/template/spec/containers/0/env/-
```
javipolo: Can we try to do strategic patch?

It will target the container by name, as coderabbit suggests, and I think it's also more robust for the env vars, in case they exist already

Maybe something like this (I did not try):

```
kubernetes.core.k8s_json_patch:
    api_version: apps/v1
    kind: Deployment
    name: osac-operator
    namespace: osac
    type: strategic
    patch:
      spec:
        template:
          spec:
            containers:
              - name: osac
                env:
                  - name: OSAC_AAP_URL
                    value: "{{ _osac_aap_url }}"
                  - name: OSAC_AAP_TOKEN
                    valueFrom:
                      secretKeyRef:
                        name: osac-aap-secret
                        key: token
```

javipolo (changes requested):

eurijon:
✨ **Claude Code**: Addressed all review comments in 7ecff61:

1. **Idempotency** — Added a pre-check that reads the current deployment and skips the patch if `OSAC_AAP_URL` and `OSAC_AAP_TOKEN` env vars already exist.

2. **Hardcoded container index** — Instead of looking up by name (the container is actually named `manager`, not `osac-operator`), added an assertion that the deployment has exactly one container. This fails clearly if sidecar injection or chart changes alter the container layout, rather than silently patching the wrong container.

3. **No-Sensitive-Data-In-Logs** — Removed AAP URL from the debug log message.

Tested end-to-end on openshift-qe-002 — full deploy passed including post-validate.

@coderabbitai approve

plugins/osac/tasks/deploy.yaml
eurijon: I saw this message after sending the second commit :smile:. I'll address it with:
```
- name: Patch OSAC operator deployment with AAP URL and AAP token
  kubernetes.core.k8s:
    state: patched
    kind: Deployment
    name: osac-operator
    namespace: osac
    definition:
      spec:
        template:
          spec:
            containers:
              - name: manager
                env:
                  - name: OSAC_AAP_URL
                    value: "{{ _osac_aap_url }}"
                  - name: OSAC_AAP_TOKEN
                    valueFrom:
                      secretKeyRef:
                        name: osac-aap-secret
                        key: token
```
setting `state: patched` which defaults to strategic merge patch and changing the container name to `manager` as the initial `osac-operator` was incorrect. I'm testing it and will send the commit soon.

eurijon:
✨ **Claude Code**: Switched to strategic merge patch in bc9bdff per @javipolo's suggestion.

`state: patched` (strategic merge) works — the earlier "image: Required value" error was from `state: present` (full apply), not from strategic merge patch itself.

This replaces the JSON Patch + pre-check approach with a single task that is:
- **Idempotent** — env vars merge by `name`, no duplicates on re-run
- **Container-name-based** — targets `manager` by name, not by index
- **Simpler** — one task instead of four

Also removed AAP URL from the debug log message (no-sensitive-data check).

Tested end-to-end on openshift-qe-002 — deploy + post-validate passed. Verified no duplicate env vars on the deployment.

plugins/osac/tasks/deploy.yaml
eurijon: Should be addressed now

javipolo (approved):

PR #517, opened 2026-06-17
fix: route logging to stderr to protect stdout tool output

rporres:
Python logging was configured to write to sys.stdout, which corrupts the machine-readable output of tools that write digest refs and PEM certificates to stdout for shell script consumption.

## Summary by CodeRabbit

* **Refactor**
  * Updated logging output handling to use the logging system instead of direct stderr writes for diagnostic messages.
  * Adjusted Ansible task output to reference error logs instead of standard output.

### DIFF
--- a/playbooks/tasks/configure_operator.yaml
+++ b/playbooks/tasks/configure_operator.yaml
@@ -151,12 +151,12 @@
       --no-dry-run
   register: operator_update_result
 
-- name: Print the output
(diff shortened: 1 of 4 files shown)


maorfr (approved):

maorfr:
this may explain https://github.com/gori-project/GoRI/issues/924

PR #526, opened 2026-06-18
fix: trust external Quay registry CA for UpdateService

danielerez:
When a Quay route is signed by an external CA (for example ZeroSSL), UpdateService failed with "unable to get local issuer certificate" because resolve-quay-registry-ca could fall back to the ingress router-ca even though that CA does not sign the live route cert.

Keep router-ca for default ingress-signed routes, but for all other chains store the intermediate and root CAs fetched from the registry. Fail loudly when the TLS chain cannot be fetched instead of writing a misleading router-ca entry into image.config additionalTrustedCA.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Quay registry certificate handling to distinguish between ingress-router-signed and externally-signed certificate paths.

* **Bug Fixes**
  * Improved registry CA certificate validation to require valid TLS chain data and extract proper certificate bundles for trust verification.

* **Tests**
  * Expanded test coverage for certificate chain handling and updated validation scenarios.

### DIFF
--- a/src/enclave/tools/quay_registry_ca.py
+++ b/src/enclave/tools/quay_registry_ca.py
@@ -110,14 +110,21 @@ def _openssl_verify(ca_pem: str, cert_pem: str) -> bool:
     return result.returncode == 0
 
 
+def _chain_trust_anchor_pem(chain: list[str]) -> str:
+    """Return CA PEM bundle from a fetched TLS chain (all certs except the leaf)."""
+    ca_certs = chain[1:]
+    if not ca_certs:
+        msg = "TLS chain contains only a leaf certificate; cannot determine trust anchor"
+        raise RuntimeError(msg)
+    return "\n".join(ca_certs) + "\n"
+
+
 def resolve_registry_ca_pem(hostname: str, *, oc: str = "oc") -> str:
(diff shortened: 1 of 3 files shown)


src/enclave/tools/quay_registry_ca.py
```diff
@@ -110,23 +110,30 @@ def _openssl_verify(ca_pem: str, cert_pem: str) -> bool:
     return result.returncode == 0
 
 
+def _chain_trust_anchor_pem(chain: list[str]) -> str:
+    """Return CA PEM bundle from a fetched TLS chain (all certs except the leaf)."""
+    ca_certs = chain[1:]
+    if not ca_certs:
+        msg = "TLS chain contains only a leaf certificate; cannot determine trust anchor"
```
maorfr:
```suggestion
        msg = (
            "TLS chain contains only a leaf certificate; cannot determine trust anchor"
        )
```
this will fix the formatting issue. don't apply this change, let's let the e2e complete first.

maorfr (approved):

maorfr:
possibly related to https://github.com/gori-project/GoRI/issues/924, answering https://github.com/gori-project/GoRI/issues/924#issuecomment-4737653400

PR #543, opened 2026-06-23
fix: isolate openssl verify from system trust store in quay registry CA check

rporres:
The router-ca verification in resolve_registry_ca_pem used openssl verify without -no-CAfile/-no-CApath, so the system CA store was also consulted. On RHEL 10, ISRG Root X1 is trusted system-wide. When the Quay TLS chain includes Let's Encrypt E8 as the first cert (non-standard server ordering), _openssl_verify(router_ca, E8) succeeded causing a false positive that made the tool return the router-ca instead of the correct Let's Encrypt chain.

Adding -no-CAfile -no-CApath restricts verification to the explicitly provided CA only, matching the intent of the router-ca check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated certificate verification for Quay registry operations to enforce use of explicitly provided CA certificates while excluding system default trust sources.

* **Tests**
  * Added test coverage to verify certificate verification behavior isolation.

### DIFF
--- a/src/enclave/tools/quay_registry_ca.py
+++ b/src/enclave/tools/quay_registry_ca.py
@@ -95,7 +95,15 @@ def _openssl_verify(ca_pem: str, cert_pem: str) -> bool:
         cert_path.write_text(cert_pem, encoding="utf-8")
         try:
             result = subprocess.run(
-                ["openssl", "verify", "-CAfile", str(ca_path), str(cert_path)],
+                [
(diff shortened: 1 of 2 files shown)


maorfr (approved):
LGTM 🤞

maorfr:
Related: https://github.com/gori-project/GoRI/issues/924

PR #548, opened 2026-06-25
Add remote plugin support

maorfr:
## Summary

Add support for remote plugins that fetch content from git repositories before deployment. This enables plugins to be maintained in separate repositories and referenced from the main enclave repository.

## Changes

- **schemas/plugin.yaml**: Add `remote` type to plugin type enum and define remote configuration schema (url, ref, path)
- **playbooks/tasks/fetch_remote_plugin.yaml**: New Ansible task that clones remote git repositories and extracts plugin content
- **playbooks/tasks/deploy_plugin.yaml**: Integrate fetch step at the beginning of plugin deployment workflow
- **scripts/verification/validate_plugins.sh**: Add 'remote' to valid plugin fields
- **plugins/example-remote/plugin.yaml**: Example remote plugin stub demonstrating the feature

## How It Works

1. Remote plugin stubs are simple `plugin.yaml` files with `type: remote` and a `remote` section
2. When deploying, the fetch task clones the specified repository (shallow clone, depth=1)
3. The content at the specified path replaces the local plugin directory
4. The actual plugin descriptor is loaded from the fetched content
5. Deployment continues normally with the fetched plugin

## Example Remote Plugin

```yaml
---
name: example-remote
type: remote

remote:
  url: https://github.com/rh-ecosystem-edge/enclave-plugin-example
  path: /example
  ref: main
```

## Test Plan

- [ ] Plugin validation passes for all existing plugins including example-remote
- [ ] Manual test: Deploy example-remote plugin to verify fetch and deployment works end-to-end
- [ ] Verify remote plugin content is correctly extracted to the local plugin directory

## Summary by CodeRabbit

* **New Features**
  * Added support for plugins that are fetched from a remote Git repository before deployment.
  * Remote plugin definitions are now validated and loaded automatically during plugin setup.

* **Bug Fixes**
  * Added checks to prevent unsafe remote paths and missing remote plugin files.
  * Improved validation so remote plugins fail fast with clearer error messages.

* **Chores**
  * Updated plugin metadata and validation rules to recognize the new remote plugin format.


rccrdpccl:
Would this automatically work in disconnected envs or for now it's only connected?

rporres:
A few things that come into my mind after a quick read

* I don't think this goes through plugin validation, and we need that to ensure plugin is healthy
* plugin vars are loaded in common/load-vars.yaml, so we may need to fetch remote plugins earlier
* I would add the addition to the plugin to CI, to make sure it works (not as a initial-installation plugin as we know it breaks but in a subsequent task)

maorfr:
> Would this automatically work in disconnected envs or for now it's only connected?

i expect it to work in disconnected.

maorfr:
## Updates Based on Review Feedback

All CodeRabbit review comments have been addressed. Here's a summary of the changes:

### Security Fixes
- ✅ **Path Traversal Prevention** (Critical): Added validation to reject absolute paths and `..` segments in `remote.path`
- ✅ **URL Credential Sanitization** (Major): Sanitize `remote_url` before logging to prevent credential leakage
- ✅ **URL Schema Restriction** (Major): Added pattern validation to only accept `https://`, `ssh://`, or `git@` URLs

### Validation Improvements
- ✅ **CI Validation** (Major): `validate_plugins.sh` now fetches and validates remote plugin content during CI, not just stubs
- ✅ **Runtime Validation** (Major): Added `ansible.utils.validate` to validate fetched `plugin.yaml` against schema
- ✅ **Schema Enforcement** (Major): Forbid `remote` field on foundation/addon plugin types

### Functional Fixes
- ✅ **Early Fetch Integration** (Major): Remote plugins now fetched before `load-vars.yaml` in both entrypoints (`bootstrap.sh` and `make deploy-plugin`)
- ✅ **Template Rendering** (Major): Template validation now fetches remote plugins before rendering
- ✅ **Validation Gate** (Major): Removed `plugin.remote is defined` check to let validation happen properly
- ✅ **Shellcheck Compliance**: Fixed SC2064 warning in `validate_plugins.sh`

### Design Decision
- **Idempotency**: Remote plugins are re-fetched on every run by design to ensure latest content from repository

All changes have been tested locally and pushed. CI builds should pass once the template rendering validation completes.

Commits since review:
- `e81076d` - Early fetch before load-vars
- `9333a32` - Security and validation fixes
- `73f0194` - CI validation for fetched content
- `ec120d2` - Fix shellcheck warning
- `9897bb6` - Fix template validation
- `db9770f` - Fix schema path resolution

maorfr:
@rporres Thanks for the review! Here's how we addressed your comments:

### 1. Plugin Validation
> "I don't think this goes through plugin validation, and we need that to ensure plugin is healthy"

**Fixed in commits `73f0194` and `9333a32`:**
- **CI validation**: `scripts/verification/validate_plugins.sh` now detects remote plugins, clones them, and validates the fetched `plugin.yaml` content (not just the stub)
- **Runtime validation**: `playbooks/tasks/fetch_remote_plugin.yaml` validates fetched plugin descriptors against `schemas/plugin.yaml` using `ansible.utils.validate`
- Both validations ensure remote plugin content is healthy before deployment/merge

### 2. Early Fetch for Plugin Vars
> "plugin vars are loaded in common/load-vars.yaml, so we may need to fetch remote plugins earlier"

**Fixed in commit `e81076d`:**
- Remote plugins are now fetched **before** `load-vars.yaml` runs in both entrypoints:
  - **Bootstrap flow** (`bootstrap.sh`): Added fetch step inside `common/load-vars.yaml` that runs before loading plugin configs
  - **Single plugin** (`make deploy-plugin`): Added fetch step in `deploy-plugin.yaml` before including `load-vars.yaml`
- This ensures plugin defaults, schemas, and configs from remote repositories are available when needed

### 3. CI Testing
> "I would add the addition to the plugin to CI, to make sure it works"

**Fixed in commit `73f0194`:**
- `validate_plugins.sh` now includes a full remote plugin fetch cycle during CI
- For each remote plugin, it:
  1. Clones the repository (shallow clone)
  2. Validates the path exists
  3. Validates the fetched `plugin.yaml` structure and fields
  4. Ensures fetched content isn't itself `type: remote` (no nested remotes)
- The `lvms` plugin (currently a remote stub for testing) successfully goes through this validation in CI

All your concerns are now addressed! 🎉

maorfr:
@coderabbitai review

plugins/lvms/plugin.yaml
```diff
@@ -1,39 +1,8 @@
 ---
 name: lvms
-type: foundation
-order: 10
+type: remote
```
rporres: can we start with an addon plugin instead? The example addon plugin would be great, something that does exactly what you want to test this and add it into .github workflow. I don't think foundation plugins should be the ones chosen to test this, given its sensitive nature and the deep catalog implications.

scripts/verification/validate_plugins.sh
```diff
@@ -80,15 +80,16 @@ for plugin_dir in "$PLUGINS_DIR"/*/; do
         continue
     fi
 
-    if ! python3 -c "
+    # Validate stub plugin.yaml and capture if it's a remote type
```
rporres: A very good time to refactor this whole spaghetti plugin validation code and write a proper python script with tests, bells and whistles

plugins/lvms/plugin.yaml
```diff
@@ -1,39 +1,8 @@
...
-    - path: "tasks/quay.yaml"
-      description: "LVMS Quay storage configuration"
-
-registries:
-  - location: "registry.redhat.io/lvms4"
-    mirror: "lvms4"
+remote:
+  url: https://github.com/rh-ecosystem-edge/enclave-plugins
+  path: lvms
+  ref: main
```
rporres: I would leave `type` untouched (the difference addon/foundation is still useful) and I would add a different way to know if we have to fetch a plugin or not through a new flag, something like `location` or something like that. With that you could even check schema as a `location: git` would have a different schema than a `location: https`.

rporres:
Thinking out loud, another way to think this in a more general way is to have a file that registers all plugins and their locations. Something that is read at the beginning of bootstrap or deploy-plugin. That list would contain something like:

```
- name: yak-shaver
  location: git
  remote:
    url: https://github.com/yak/shaver
    dir: /
    ref: main

  name: clair-import
  location: git
  remote:
    url: https://github.com/rh-ecosystem-edge/enclave
    dir: /plugins/clair-import
    ref: main
```
 
With that you wouldn't need to touch the schema at all and there are no differences between a remote plugin and a local enclave plugin. You have something at the beginning of your run that fetches what you need.

I would also think in a way of supporting unregistered plugins in some way, in case an external plugin is needed. But it would follow the same mechanism and no hacks would be needed.

plugins/lvms/plugin.yaml
maorfr: yes, just used lvms to prove the approach works.

maorfr:
at this point all the tests are passing, which proves feasibility.

will foĺlow up next week!

scripts/verification/validate_plugins.sh
rporres: Another thought about refactoring this would be the possibility to run it in the plugins repos to have early feedback of what's going to be validated when a plugin is used in enclave.

maorfr:
## CodeRabbit Review Comments Addressed

All actionable CodeRabbit review comments have been resolved:

### 1. ✅ Only Fetch Enabled Plugins (Major - Stability)
**Issue**: Remote plugin fetch ran for ALL plugins, not just enabled ones, causing failures on disabled plugins.

**Fix** (commit `e534725`): Added check in `playbooks/common/load-vars.yaml` to only fetch plugins in `enabled_plugins` list:
```yaml
when:
  - item.item.path | basename in enabled_plugins
  - vars['plugin_descriptor_' + basename].remote is defined
```

### 2. ✅ Double-Fetch Performance Issue (Minor)
**Issue**: Concern about fetching same plugin twice (deploy-plugin.yaml and load-vars.yaml).

**Analysis**: Not actually an issue - after first fetch, the plugin directory is replaced with fetched content which doesn't have a `remote` field, so the second check `plugin.remote is defined` will be false.

### 3. ✅ Python Syntax Error & Path Injection (Critical/Major - Security)
**Issue**: Complex bash-embedded Python in `validate_plugins.sh` had syntax errors and path injection vulnerabilities.

**Fix** (commit `7bd828a`): Simplified approach - removed all complex validation logic, only added 'remote' to valid fields list. Defers full remote plugin validation to future PR when the script is refactored to Python (as suggested by @rporres).

### 4. ✅ Mutable Branch Security Concern (Major)
**Issue**: Using `ref: main` is mutable and can change without review.

**Fix** (commit `e08b974`): Changed clair-import to use immutable commit SHA instead of `main`:
```yaml
ref: 78200c11c0a18953cec0ee45f4f1a0bbb46b75de
```

### Architecture Changes

Also implemented @rporres's feedback to use `remote` field instead of `type: remote`:
- Commit `17d95fd`: Changed from `type: remote` to using `remote` field on existing foundation/addon types
- Commit `68eefa7`: Restored lvms to local plugin (was only for testing)
- Commit `4193d54`: Converted clair-import to remote plugin as demonstration

This keeps the `type` field focused on deployment behavior (foundation vs addon) and treats remote location as an orthogonal concern.

---

All commits since last review:
- `e08b974` - Use immutable commit SHA for clair-import
- `e534725` - Only fetch enabled remote plugins
- `7bd828a` - Simplify validate_plugins.sh 
- `4193d54` - Convert clair-import to remote plugin
- `68eefa7` - Restore lvms plugin
- `17d95fd` - Change to use remote field instead of type:remote
- `db9770f` - Fix schema path resolution
- `9897bb6` - Fix template validation
- `ec120d2` - Fix shellcheck warning
- `73f0194` - Add CI validation

All CI checks should now pass. Ready for re-review.

plugins/lvms/plugin.yaml
maorfr: switched to clair-import (which already has a dependency on github as is)

scripts/verification/validate_plugins.sh
maorfr: reverted all changes and left the refactor for a future PR

plugins/lvms/plugin.yaml
maorfr: left `type` untouched and only using `remote`.

maorfr:
> Thinking out loud, another way to think this in a more general way is to have a file that registers all plugins and their locations. Something that is read at the beginning of bootstrap or deploy-plugin.

this sounds appropriate for `enabled_plugins`. i do like the current simplicity of using plugin names as a unique key and reading content from plugin files. leaving things as is for now.

maorfr:
@coderabbitai re-review please

scripts/verification/validate_plugins.sh
rporres: Why waiting? We have time to do this properly in the beginning of this development cycle

rporres:
> this sounds appropriate for enabled_plugins. i do like the current simplicity of using plugin names as a unique key and reading content from plugin files. leaving things as is for now.

It is very different to that. What I'm proposing is a unified way of dealing with plugins no matter where they are stored or how they are served. The list would help as a register, a way to know how to add a enabled_plugin at bootstrap or how do via a deploy-plugin: it has a way to fetch the content (git, https, etc), and once the content is fetched every plugin goes through the same steps. 

I realise it is not very different to a helm repo index

I would like something for plugins that is as future proof as we can imagine and that can be used by external parties without any modification in the repo needed, e.g. being able to give a custom index file or the location of the plugin.

PR #554, opened 2026-06-28
Fix virtual media detach timing to prevent ISO reboot loop

maorfr:
## Summary

Fixes https://github.com/gori-project/GoRI/issues/937

During Agent-Based Installer (ABI) deployments, hosts were booting back into the virtual media ISO instead of from disk after the installer wrote OpenShift and triggered a reboot, causing the deployment to block with `installing-pending-user-action`.

## Root Cause

The virtual media detach task was executed too late in the deployment flow - **after** waiting for `install-complete` in `wait_for_deployment.yaml`. The ABI installer reboots hosts mid-installation (after writing OS to disk) but before reaching `install-complete`, so virtual media was still attached during those critical mid-install reboots.

On systems where UEFI firmware prioritizes virtual media over disk, hosts would boot back into the ISO instead of from disk.

## Changes

- **`playbooks/03-deploy.yaml`**: Added virtual media detach immediately after all hosts reach Ironic `active` state and before `wait_for_deployment`
- **`playbooks/tasks/wait_for_deployment.yaml`**: Removed the duplicate (and too-late) detach task

## Impact

Virtual media is now detached as soon as Ironic confirms nodes are active (ISO successfully booted), ensuring the firmware's virtual media entry is removed before any installer-triggered reboots occur. All subsequent reboots will land on disk as expected.

## Testing

This fix addresses the exact sequence described in GoRI #937 and follows the recommended solution from that issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Virtual media detachment is now executed explicitly immediately after hosts reach an active state in the deployment workflow to keep systems clean.
  * The deployment wait phase no longer performs additional virtual media detachment, simplifying the transition from installation completion to cleanup.
  * Virtual media detachment now treats “not found” (HTTP 404) as non-fatal, while still failing on unexpected responses.

### DIFF
--- a/playbooks/03-deploy.yaml
+++ b/playbooks/03-deploy.yaml
@@ -75,6 +75,7 @@
           loop: "{{ agent_hosts }}"
           loop_control:
             loop_var: agent_host
+            label: "{{ agent_host.name }}"
           ansible.builtin.include_tasks:
             file: tasks/configure_hardware_ironic_detach_vmedia.yaml
             apply:
--- a/playbooks/03-deploy.yaml
+++ b/playbooks/03-deploy.yaml
@@ -71,6 +71,16 @@
               tags: hardware
           tags: hardware
 
+        - name: Detach virtual media from hosts
+          loop: "{{ agent_hosts }}"
+          loop_control:
+            loop_var: agent_host
+          ansible.builtin.include_tasks:
+            file: tasks/configure_hardware_ironic_detach_vmedia.yaml
+            apply:
+              tags: hardware
+          tags: hardware
+
         - name: Pre-install validate plugins
           ansible.builtin.include_tasks:
             file: tasks/pre_install_validate_plugins.yaml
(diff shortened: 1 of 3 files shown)


eurijon (approved):

rporres (approved):
apart from the coderabbitai comment, it looks fine :)

maorfr:
✨ **Claude Code**: Addressed CodeRabbit review feedback in commit 914da75:

1. **No-Sensitive-Data-In-Logs**: Added `loop_control.label: "{{ agent_host.name }}"` to prevent logging sensitive host data (redfishPassword, BMC details)

2. **Error handling**: Replaced broad `ignore_errors: true` with explicit status code handling:
   - Accept `204` (successfully detached) and `404` (already detached/not found) as success
   - Fail on unexpected errors (401 auth failures, 500 server errors, timeouts)
   - This ensures real Ironic failures propagate instead of being silently ignored

@coderabbitai ready for re-review

maorfr:
✨ **Claude Code**: Fixed template error handling in commit d4b2421:

Added `| default(0)` to the `failed_when` condition to safely handle cases where the URI call fails before setting the status field (e.g., connection timeout, DNS failure).

This prevents Jinja2 template errors while still failing the task on unexpected HTTP status codes.

@coderabbitai approve

maorfr:
@coderabbitai review

PR #555, opened 2026-06-30
OSAC-1535: Fix E2E OSAC workflow startup_failure

eurijon:
## Summary
- Add `checks: write` permission to `e2e-osac.yml` workflow
- The reusable `e2e-deployment.yml` requires `checks: write` to post check results, but `e2e-osac.yml` only granted `contents: read`, causing `startup_failure` on every run since the workflow was created

## Test plan
- [x] Verify E2E OSAC workflow no longer fails with `startup_failure`
- [x] Confirm the workflow runs end-to-end when triggered by plugin changes

## Summary by CodeRabbit

* **Chores**
  * Improved the reliability of end-to-end validation by allowing test runs to update check statuses properly during connected OSAC execution.

### DIFF
--- a/.github/workflows/e2e-osac.yml
+++ b/.github/workflows/e2e-osac.yml
@@ -28,15 +28,15 @@ on:
       - 'plugins/osac/**'
       - '.github/workflows/e2e-osac.yml'
 
-permissions:
-  contents: read
-
 concurrency:
   group: e2e-osac-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
 jobs:
   e2e-osac:
+    permissions:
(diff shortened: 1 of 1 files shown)


eurijon:
@coderabbitai review

.github/workflows/e2e-osac.yml
rporres: nit: is this still needed?

rporres (approved):

.github/workflows/e2e-osac.yml
eurijon: Removed. The job-level permissions block already sets contents: read + checks: write, so the top-level one was redundant

PR #565, opened 2026-07-02
rpm generation

rccrdpccl:
cc @javipolo @maorfr 

we could build disconnected RPM (with uv dependencies cached within)) or connected.

This an example on how it would work, if we like the idea we can improve it further


maorfr:
mind moving from hack to scripts? so it's not included in the tarball.

maorfr:
while we are here, should we also add a github workflow to build and publish the rpm? possibly similar to the tarball, as an OCI artifact in quay.io? or is there a better alternative?

hack/rpm/build-rpm.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# --- Step 1: Create source tarball ---
+echo "[1/4] Creating source tarball..."
+git -C "${REPO_DIR}" archive \
+    --format=tar.gz \
+    --prefix="enclave-${VERSION}/" \
+    --output="${WORK_DIR}/enclave-${VERSION}.tar.gz" \
+    HEAD
+
+# --- Step 2: Augment tarball for disconnected mode ---
```
javipolo: I dont understand this. Why to have different tarballs for connected and disconnected?

We assume LZ is always connected, so embedding everything should not be needed
However, if we want to embed everything, we should always do it, dont we?

hack/rpm/enclave.spec
```diff
@@ -0,0 +1,128 @@
...
+    target="${f%.example.yaml}.yaml"
+    [ -f "${target}" ] || cp "${f}" "${target}"
+done
+for f in "${ENCLAVE_DIR}/config/plugins/"*.example.yaml; do
+    [ -f "${f}" ] || continue
+    target="${f%.example.yaml}.yaml"
+    [ -f "${target}" ] || cp "${f}" "${target}"
+done
+
+%if "%{enclave_mode}" == "connected"
```
javipolo: Should we do all this here or we just let bootstrap do it?

rccrdpccl: you're right, probably bootstrap should do it

hack/rpm/build-rpm.sh
rccrdpccl: I've assumed it would help, however it would generate a huge RPM (~500MB). hence I thought dual mode might help. in any case let's start without disconnected mode and we'll take it from there

rccrdpccl:
> while we are here, should we also add a github workflow to build and publish the rpm? possibly similar to the tarball, as an OCI artifact in quay.io? or is there a better alternative?

I would just attach it to the release, definitely we'll add a github workflow to test/dry-run and publish - I'll add it before wrapping up, in the case we change approach during the review

PR #570, opened 2026-07-07
OSAC-584: Make hardware configuration step idempotent

maorfr:
## Summary
- Rerunning Phase 3 (`playbooks/03-deploy.yaml`) after a cluster is already installed — or resuming an interrupted bootstrap — currently redeploys Ironic and reboots every `agent_host` via virtual media, even ones already serving as active cluster nodes (Ironic is torn down after every successful install, so a rerun recreates it from scratch with no cluster awareness).
- Adds `playbooks/tasks/configure_hardware_filter_hosts.yaml`, which queries the cluster's `Node` objects via the existing kubeconfig and computes `agent_hosts_to_provision` — the subset of `agent_hosts` that are not already `Ready` in the cluster.
- `03-deploy.yaml` now runs this filter before "Setup Ironic", skips deploying Ironic entirely when nothing needs provisioning, and loops the boot/wait/detach-vmedia steps over `agent_hosts_to_provision` instead of the full `agent_hosts` list.
- NotReady/missing nodes are never filtered out, so a broken or partially-bootstrapped host is still (re)provisioned. If the cluster API isn't reachable at all (fresh bootstrap), the check fails open and all hosts are provisioned as before — no behavior change for first-time installs.
- Per-node Ironic-level idempotency (`configure_hardware_ironic_boot/_wait/_detach_vmedia.yaml`) is untouched.

Ref: [OSAC-584](https://redhat.atlassian.net/browse/OSAC-584) (Enclave bootstrap idempotency)

## Test plan
- [x] `ansible-playbook playbooks/03-deploy.yaml --syntax-check`
- [x] `yamllint` / `ansible-lint` clean on both changed files
- [ ] End-to-end verification against a live/local cluster (e.g. sushy-tools): confirm a rerun after a completed install skips already-Ready hosts and does not reboot them, while a NotReady/missing host is still reprovisioned

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hardware provisioning to be safe to rerun without repeating work for hosts that are already ready.
  * Skips already-available agent hosts and only provisions the remaining ones.
  * Added clearer handling so follow-up setup steps run only when there are hosts left to provision.

### DIFF
--- a/playbooks/03-deploy.yaml
+++ b/playbooks/03-deploy.yaml
@@ -9,6 +9,7 @@
 # - Copy mirror manifests (disconnected only)
 # - Generate Agent-Based Installer ISO
 # - Copy ISO to web server
+# - Skip hosts already Ready in the cluster (idempotent reruns)
 # - Setup Ironic and boot agent hosts via virtual media
 # - Wait for deployment to complete
 # - Detach virtual media from hosts
@@ -44,15 +45,23 @@
               tags: configure-abi
           tags: configure-abi
 
+        - name: Filter agent hosts already provisioned in the cluster
+          ansible.builtin.include_tasks:
+            file: tasks/configure_hardware_filter_hosts.yaml
+            apply:
+              tags: hardware
+          tags: hardware
+
         - name: Setup Ironic for hardware configuration
           ansible.builtin.include_tasks:
             file: tasks/configure_hardware_ironic_setup.yaml
             apply:
               tags: hardware
           tags: hardware
+          when: agent_hosts_to_provision | length > 0
 
         - name: Configure and boot hosts via Ironic
-          loop: "{{ agent_hosts }}"
+          loop: "{{ agent_hosts_to_provision }}"
           loop_control:
             loop_var: agent_host
           ansible.builtin.include_tasks:
@@ -62,7 +71,7 @@
           tags: hardware
 
         - name: Wait for all hosts to reach active state
-          loop: "{{ agent_hosts }}"
+          loop: "{{ agent_hosts_to_provision }}"
           loop_control:
             loop_var: agent_host
           ansible.builtin.include_tasks:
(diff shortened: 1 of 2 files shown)


maorfr:
✨ **Claude Code**: Addressed both review items in e19c0a2.

**1. Fail-open on k8s_info errors (actionable comment on `configure_hardware_filter_hosts.yaml:5-13`)**

Fixed. The task now first checks whether `{{ workingDir }}/ocp-cluster/auth/kubeconfig` exists at all via `ansible.builtin.stat`:
- If it doesn't exist, this is unambiguously a fresh install with no cluster yet — proceed to provision all `agent_hosts` immediately, no retry needed.
- If it does exist, the `k8s_info` Node query now retries using this repo's standard `k8s_retries`/`k8s_delay` convention (same as the existing kubeconfig/Node checks in `wait_for_deployment.yaml`) instead of a single one-shot attempt, so a transient reachability blip on an already-live cluster gets a real chance to resolve before we conclude anything. If it's still unreachable after those retries, we fall back to provisioning all hosts, but now with an explicit debug warning ("Kubeconfig exists but cluster Node state could not be determined...") instead of a silent fallback, so operators aren't surprised by an unexpected reprovision.

**2. No-Sensitive-Data-In-Logs pre-merge check (hostnames in debug/loop labels)**

Not changing this — the flagged pattern (logging `agent_host.name` / `item.metadata.name` in debug messages and loop labels) is pre-existing and consistent throughout this exact area of the codebase, e.g. `configure_hardware_ironic_boot.yaml`'s "Display boot status" task (`"Node {{ agent_host.name }} ({{ node_uuid }}) boot initiated"`) and the per-host debug message in `configure_hardware_ironic_wait.yaml`. These are operator-chosen management-plane hostnames (e.g. `mgmt-ctl01`), not credentials or secrets, and this PR doesn't introduce a new category of log output — it follows the same convention already used by the surrounding Ironic task files it's wired alongside. Happy to redact/anonymize across the board in a follow-up if the team wants a stricter policy going forward, but that would be a broader, separate change rather than something scoped to this PR.

maorfr:
@coderabbitai review/approve

maorfr:
@coderabbitai approve

javipolo (approved):

PR #574, opened 2026-07-07
[stable] Add trust quay ca in update service migration (backport #549)

maorfr:
it is needed to make sure that update service works fine by adding the needed root CA to the trust store configmap it mounts


maorfr:
IBM code does similar operation at later deploy time

PR #584, opened 2026-07-08
OSAC-2255: Embed enclave CLI calls into upgrade playbook

maorfr:
## Summary
- Automate the management cluster and operator version upgrades (`enclave reconcile mgmt-cluster-version --use-defaults` / `enclave reconcile operator-versions --use-defaults`) as tasks in `playbooks/upgrade.yaml`, so `./upgrade.sh` fully covers steps 7-9 of the upgrade process instead of requiring separate manual commands
- Each upgrade step can be individually skipped via `-e upgrade_mgmt_cluster=false` / `-e upgrade_operators=false` for cases where timing needs to be controlled manually
- Update `docs/UPGRADE.md` to describe the automated flow and the opt-out flags

## Test plan
- [x] `ansible-lint playbooks/upgrade.yaml` passes
- [ ] Run `./upgrade.sh` against a real deployment to confirm the reconcile calls execute successfully end-to-end

## Summary by CodeRabbit

* **New Features**
  * Streamlined upgrade flow now runs migrations and then reconciles both the management cluster version and operator versions as part of `./upgrade.sh`.
  * Added an option to run management cluster and/or operator upgrades separately by invoking `ansible-playbook` with flags (e.g., `upgrade_mgmt_cluster=false` and/or `upgrade_operators=false`).

* **Documentation**
  * Updated the upgrade guide to reframe and renumber steps, clarifying how each action is executed during the upgrade process.

### DIFF
--- a/playbooks/upgrade.yaml
+++ b/playbooks/upgrade.yaml
@@ -29,9 +29,6 @@
           - reconcile
           - mgmt-cluster-version
           - --use-defaults
-          - --no-dry-run
-      environment:
-        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
       register: r_mgmt_cluster_upgrade
       when: upgrade_mgmt_cluster | default(true) | bool
 
@@ -47,9 +44,6 @@
           - reconcile
           - operator-versions
           - --use-defaults
-          - --no-dry-run
-      environment:
-        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
       register: r_operator_versions_upgrade
       when: upgrade_operators | default(true) | bool
 
--- a/playbooks/upgrade.yaml
+++ b/playbooks/upgrade.yaml
@@ -21,3 +21,39 @@
     - name: Run migrations
       ansible.builtin.include_tasks:
         file: tasks/migrations.yaml
+
+    - name: Upgrade management cluster OpenShift version
+      ansible.builtin.command:
+        argv:
+          - enclave
+          - reconcile
+          - mgmt-cluster-version
+          - --use-defaults
+          - --no-dry-run
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
+      register: r_mgmt_cluster_upgrade
+      when: upgrade_mgmt_cluster | default(true) | bool
+
+    - name: Print management cluster upgrade output
+      ansible.builtin.debug:
+        var: r_mgmt_cluster_upgrade.stderr_lines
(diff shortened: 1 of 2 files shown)


playbooks/upgrade.yaml
```diff
@@ -21,3 +21,39 @@
...
         file: tasks/migrations.yaml
+
+    - name: Upgrade management cluster OpenShift version
+      ansible.builtin.command:
+        argv:
+          - enclave
+          - reconcile
+          - mgmt-cluster-version
+          - --use-defaults
+          - --no-dry-run
```
rporres: this option shouldn't be needed, it's the default.

playbooks/upgrade.yaml
```diff
@@ -21,3 +21,39 @@
...
+      when: upgrade_mgmt_cluster | default(true) | bool
+
+    - name: Upgrade operator versions
+      ansible.builtin.command:
+        argv:
+          - enclave
+          - reconcile
+          - operator-versions
+          - --use-defaults
+          - --no-dry-run
```
rporres: ditto

playbooks/upgrade.yaml
```diff
@@ -21,3 +21,39 @@
...
+    - name: Upgrade management cluster OpenShift version
+      ansible.builtin.command:
+        argv:
+          - enclave
+          - reconcile
+          - mgmt-cluster-version
+          - --use-defaults
+          - --no-dry-run
+      environment:
+        KUBECONFIG: "{{ workingDir }}/ocp-cluster/auth/kubeconfig"
```
rporres: you can skip setting this if you want, too :)

playbooks/upgrade.yaml
maorfr: ✨ **Claude Code**: Confirmed — `dry_run` already defaults to `False` in `src/enclave/reconcile/cli.py`, so `--no-dry-run` was redundant. Removed in 826d76b.

playbooks/upgrade.yaml
maorfr: ✨ **Claude Code**: Same as above — removed in 826d76b.

playbooks/upgrade.yaml
maorfr: ✨ **Claude Code**: Confirmed — `playbooks/tasks/migrations.yaml` already symlinks `~/.config/enclave/kubeconfig`, which the CLI auto-discovers (same pattern as `playbooks/tasks/configure_operator.yaml`). Removed the explicit `KUBECONFIG` env from both tasks in 826d76b.

rporres (approved):

PR #598, opened 2026-07-13
OSAC-2322: Implement rails-style versioned migrations

rporres:
Implements the design from docs/design/rails-style-migrations.md.

Replaces the unversioned migration runner with a timestamp-based system that tracks applied migrations in an append-only control file on the LZ host (~/.config/enclave/migration_tasks), so each migration runs exactly once regardless of how many times upgrade.sh is executed.

- Rename all existing migration files with UTC timestamp prefixes and move their when: conditions inside using Ansible block: pattern
- Promote trust_quay_registry_ca_for_image_config.yaml to a timestamped migration; move operator_catalog_source.yaml to playbooks/tasks/ as a shared utility (not a migration)
- Rewrite migrations.yaml as a set-subtraction runner; add run_single_migration.yaml for fail-fast per-migration recording
- Add step_write_migration_tasks to bootstrap.sh to seed the control file on fresh installs, preventing re-execution of bundled migrations
- Add validate-migrations CI target and validate_migrations.sh script (monotonic ordering, not-in-future, uniqueness, immutability checks)
- Wire write-migration-tasks step into e2e-deployment.yml for both connected and disconnected jobs
- Document the migration system in docs/UPGRADE.md

## Summary by CodeRabbit

* **New Features**
  * Added versioned migration handling for upgrades, including automatic tracking, ordering, retrying failed migrations, and initialization on fresh deployments.
  * Added support for configuring the migration control file location.
  * Added migration steps for plugin catalogs, certificate trust, and CLI kubeconfig setup.

* **Documentation**
  * Documented migration behavior, configuration, upgrade scenarios, and migration file requirements.

* **Bug Fixes**
  * Improved deployment reporting so failures during migration setup are identified correctly.

* **Validation**
  * Added checks for migration timestamps, uniqueness, ordering, and immutability.


rporres:
✨ **Claude Code**: Addressed in 9babc4c. Summary of each finding:

**Fixed:**
- **`deploy_bootstrap_step.sh`** — added `write-migration-tasks` to the step allowlist and usage messages; without this the `deploy-cluster-write-migration-tasks` make target would have rejected the step before reaching `bootstrap.sh`.
- **`migrations.yaml`** — corrected `{{ playbook_dir }}/../tasks/migrations` to `{{ playbook_dir }}/tasks/migrations`; the `../` caused the path to resolve outside the playbook tree entirely.
- **`20260615201605_foundation_plugin_fleet_catalogs.yaml`** — corrected `include_tasks` path from `tasks/configure_operators_fleet.yaml` to `../configure_operators_fleet.yaml`; Ansible resolves `include_tasks` paths relative to the task file's directory, not the playbook directory, so the original path would never have found the file.
- **`validate_migrations.sh`** — tightened the `find` pattern from the loose `[0-9]*_*.yaml` to the strict 14-digit `[0-9][0-9]...[0-9]_*.yaml`, consistent with every other pattern in the script.

**Not addressed — `20260517122429_trust_quay_registry_ca_for_image_config.yaml`:**

Both comments (replace `ansible.builtin.shell` with `k8s_info`, tighten `failed_when: false` on the UpdateService restart) are valid improvements, but this PR is a rename/promotion of the existing `trust_quay_registry_ca_for_image_config.yaml` task file — the content is intentionally unchanged from the file that is already deployed and working in production. Changing the logic here would mix a refactor with a behaviour change inside a migration file that, by the system's own immutability rules, should not be modified after merging. A follow-up migration or a separate PR is the right place for those improvements.

@coderabbitai please review

playbooks/tasks/migrations.yaml
```diff
@@ -1,33 +1,50 @@
...
+#
+# Called from upgrade.yaml via include_tasks.
 
-- name: Foundation plugin catalog source migration
-  ansible.builtin.include_tasks:
-    file: migrations/foundation_plugin_catalog_sources.yaml
-  when: disconnected | default(true) | bool
+- name: Read LZ migration control file
+  ansible.builtin.slurp:
+    src: "~/.config/enclave/migration_tasks"
```
javipolo: Maybe we need to use the session/workingDir as the place to store the migrations control file

rporres: I think it makes sense to avoid scattering the stuff we create and keep everything under a place that it is important to keep. working dir has important stuff like the kubeconfig, so it's nice to keep new stuff under it. I will also review the documentation to make sure it conveys the idea of workingDir being something important to keep.

rporres: fixed in https://github.com/rh-ecosystem-edge/enclave/pull/598/commits/cb28c2401e89098a344bbc1def9f01af438ebd45

rporres:
✨ **Claude Code**: Addressed comments from review #4696268510 in caabb58.

@coderabbitai review

**Fixed (comments 2, 4, 5):**

- **`playbooks/tasks/migrations.yaml`** — replaced `ignore_errors: true` on the slurp with an `ansible.builtin.stat` check first. The slurp now only runs when the file exists, so real permission/IO errors are no longer swallowed silently. `_applied_migrations` is initialised as empty when the file is absent.

- **`scripts/verification/validate_migrations.sh`** — sourced `scripts/lib/output.sh` and replaced all `echo "ERROR: ..."` calls with the `error` helper for output consistency with the other validation scripts.

- **`scripts/verification/validate_migrations.sh`** — replaced the word-splitting `for fname in $modified` loop in the immutability check with a `while IFS= read -r fname` loop reading from process substitution, eliminating the intermediate variable and the word-splitting risk.

**Not fixed (comments 1, 3):**

- **Comment 1 (`bootstrap.sh` — quote `${log}`)**: `${log}` is unquoted throughout the entire file in every step function (`tee -a ${log}` appears ~10 times). Quoting it only in `step_write_migration_tasks` would be inconsistent. The value is constructed from `date +%Y%m%d_%H%M%S` so spaces are impossible in practice.

- **Comment 3 (`operator_catalog_source.yaml` — `spec.source` undefined)**: In Ansible/Jinja2 the undefined-key comparison on line 51 evaluates to `False` and falls through safely. A Kubernetes `Subscription` object always has `spec.source` set as a required field, so this code path cannot be reached in practice.

rporres:
@coderabbitai please approve

maorfr (approved):
LGTM

beautiful 😽

PR #599, opened 2026-07-14
docs: note RHCOS ISO extraction can be skipped by pre-placing a file

maorfr:
## Summary
- Documents the existing skip-if-exists behavior for RHCOS ISO extraction: if a file already exists at the expected destination path for a given version, the prepare phase skips extracting it from the release image
- Clarifies that this allows users to supply their own ISO by pre-placing it at that path

Code reference: https://github.com/rh-ecosystem-edge/enclave/blob/6cf151e0fe91fdcbfe04bf2a77b4c3a73b56f4f6/playbooks/tasks/download_content_single_version.yaml#L2-L5

## Test plan
- [x] Docs-only change, no code affected

## Summary by CodeRabbit

* **Documentation**
  * Clarified RHCOS ISO handling during preparation.
  * Documented that existing ISOs are reused, and custom ISOs should be placed at the expected destination before preparation.

### DIFF
--- a/docs/CONFIGURATION_REFERENCE.md
+++ b/docs/CONFIGURATION_REFERENCE.md
@@ -1356,6 +1356,12 @@ RHCOS ISOs are automatically extracted from OpenShift release images during the 
 3. Determines the ISO volume ID using `blkid`
 4. Stores the ISO at `/var/www/html/rhcos-<version>-x86_64-live-iso.x86_64.iso`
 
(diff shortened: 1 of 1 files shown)


maorfr:
Related to gori-project/GoRI#941 — documents an existing mechanism (skip-if-exists) that partially addresses the request there for a way to bypass ISO extraction.

eurijon (approved):

PR #605, opened 2026-07-16
Add shared knowledge base at .claude/knowledge/

maorfr:
## Summary
- Introduce `.claude/knowledge/`, a git-checked-in place for non-obvious project facts, conventions, gotchas, and settled decisions mined from past Claude Code sessions — shared across everyone's Claude Code instance working in this repo, so nobody has to re-spend tokens re-deriving the same investigations.
- Add `AGENTS.md` guidance pointing to both `docs/` (formal deployment/architecture/config guides) and `.claude/knowledge/` (informal knowledge) as places to check before investigating something, and to update after learning something worth sharing.
- Seed the knowledge base with 17 files covering recurring CodeRabbit/CI patterns, architecture gotchas (InfraEnv readiness races, Ironic lifecycle, UpdateService CA trust, catalog naming, schema drift, etc.), a security note (never `--debug` the `jira` CLI), and a couple of PR-scoped settled decisions.

## Test plan
- [x] `README.md` documents the format and inclusion/exclusion criteria for future entries
- [ ] Reviewer spot-checks a few entries for accuracy against current `main`


eranco74:
Like the concept — shared, git-checked-in knowledge that saves every contributor's Claude instance from re-deriving the same non-obvious facts is a real token saver. The seed entries are well-structured and clearly actionable. A few risks worth thinking about:

**1. Staleness is the biggest risk.** Entries reference specific files, line numbers, and PR numbers. As the codebase evolves, these become stale or wrong — and stale entries are worse than no entries because they give false confidence. Consider adding a lightweight staleness check: a CI job or pre-commit hook that verifies file paths referenced in knowledge entries still exist (most entries reference specific files like `playbooks/tasks/configure_operator.yaml` or `schemas/plugin.yaml`). Broken paths are the cheapest signal that an entry needs review.

**2. Hallucination laundering.** These entries were mined from past Claude Code sessions. If Claude misunderstood something during a session, that misunderstanding now gets committed to git and treated as authoritative by future Claude instances. The "spot-check a few entries" test plan item is good but 17 files with deep Ansible/Ironic/Metal3 details is a lot to verify in one pass. Consider requiring that each new knowledge entry includes a `verified-by:` field in the frontmatter with a human reviewer's name — this makes the accuracy bar explicit and traceable.

**3. Overlap with AGENTS.md.** `config-schema-sync.md` acknowledges "This is already called out in AGENTS.md" but exists anyway. Having the same rule in two places means two places to maintain and eventually disagree. Suggest a clear boundary: AGENTS.md owns the rules, `.claude/knowledge/` owns the *why behind the rule and how it was missed in practice*. If a knowledge entry's "How to apply" duplicates an AGENTS.md rule, it should reference the AGENTS.md section instead of restating it.

**4. PR-scoped entries need a cleanup mechanism.** `remote-plugin-design-decisions.md` says "update or remove once the PR merges" — but who will actually do that? Consider tagging these with `scope: pr-548` in frontmatter and adding a simple grep to the PR merge checklist or a bot that flags entries whose scoped PR has merged/closed.

**5. Claude-specific path.** `.claude/knowledge/` won't be auto-discovered by Cursor, Copilot, or Gemini. The AGENTS.md addition helps, but you could also symlink or reference it from `.cursor/` and `.gemini/` directories if those tools are used in this repo.

None of these are blockers — the content quality is high and the value is clear. Just want to make sure the knowledge base stays trustworthy as it grows.

rccrdpccl:
nice idea @maorfr ! How is the agent discovering this info? Would it be worth to keep an index file with 1 sentence description for each chapter for easier discovery?

maorfr:
✨ **Claude Code**: Addressed in 63a11ba.

@eranco74 thanks, all fair points:
1. **Staleness** — stripped line numbers, exact "currently X/Y/Z" lists, and a stale-prone unmerged-branch reference from several entries; the README now explicitly instructs future entries to describe patterns/how-to-verify rather than exact current state, and to fix or delete an entry (as its own small PR if unrelated to current work) on noticing it's gone stale.
2. **Hallucination laundering** — added a README section telling future Claude instances (and reviewers) to treat entries with some skepticism, spot-check load-bearing claims against current code, and periodically re-challenge entries rather than trusting them at face value. Didn't add a `verified-by:` field — didn't want to imply a human has signed off on each one when that's not really enforced; the skepticism instruction felt more honest about what this actually is.
3. **Overlap with AGENTS.md** — `config-schema-sync.md` now defers to AGENTS.md for the rule itself and only keeps the "why it's still missed in practice despite being documented" part.
4. **PR-scoped entries** — removed `remote-plugin-design-decisions.md` entirely rather than adding a cleanup mechanism; agreed there's no reliable "who comes back to remove it."
5. **Tool-specific path** — added a note in AGENTS.md asking contributors using non-Claude tools to request a reference/symlink rather than duplicating content.

@rccrdpccl re: discovery — a Claude session mined its own past session transcripts (grepping for review-comment/decision signal) for candidates, then I picked which to write. Added the index you asked for to README.md, now that the entries themselves are less stale-prone.

maorfr:
@coderabbitai review

.claude/knowledge/tls-cert-test-coverage-gap.md
```diff
@@ -0,0 +1,16 @@
+---
```
rporres, line 1: this one is problematic, since I'm already working on a fix.

I would not add it and in general prevent from including known gaps since they're good candidates for becoming stale in a short time.

.claude/knowledge/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Statement of the fact/rule/decision, then:
+
+**Why:** the reasoning or incident that established it.
+**How to apply:** what this should change about future work.
+```
+
+Add a one-line entry to the index below when you add a file, and remove it when you delete one.
+
+## Index
```
rporres, line 72: we should have a ci task to make sure this is up to date and make sure that knowledge files follow the expected schema.

.claude/knowledge/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Treat entries with some skepticism
+
+Every entry here was written by a Claude instance, generally by mining past session
+transcripts — not by a human independently verifying the claim. That means a misunderstanding
+from one session can get "laundered" into something that reads as settled, citable fact. Before
+relying on an entry to make a decision (not just as a lead to investigate further), spot-check
+its concrete claims against the current code rather than taking it on faith — especially for
+anything load-bearing. Periodically challenge entries you're using: does this still hold, or
+was it true once and never revisited?
```
rporres, line 39: "some skepticism" looks dangerous. Entries should be either valid or wrong and we have to a process to remove them if they're not right anymore. A wrong entry is worse than a missing entry.

.claude/knowledge/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Entries describe the codebase as it was when written, not as it will always be. Prefer
+referencing files and describing patterns/behavior over citing specific line numbers, exact
+current lists (e.g. "which plugins currently do X"), or the live/merged state of a specific
+open PR or branch — all of those drift out of date fast, and a wrong specific is worse than a
+vague one because it reads as authoritative. When you notice an entry no longer matches the
+code (a referenced file/behavior has changed, a cited PR merged and changed the picture, a
+"known gap" got closed), don't just ignore it — fix or delete the entry. If that's unrelated to
+what you're currently working on, do it as a small separate PR rather than folding it into an
+unrelated change.
```
rporres, line 51: I'm always very skeptical of good intentions... We may want to rewrite this a set of rules regarding what constitutes a good knowledge entry so that they're read by agents and they can detect when a knowledge entry is no longer valid since it's being addressed.

rporres (commented):
I like this idea. Some of the entries really have an ADR feeling, but it's fine as they are since we really don't have processes around new features.

Staleness is the thing that worries me the most.

eranco74 (commented):
Overall I like this — having a shared knowledge base that saves everyone from re-deriving the same investigations is a solid idea.

Consider addressing the following before merging:

1. **Drop transient entries like `tls-cert-test-coverage-gap.md`** — known gaps and bugs-in-progress belong in Jira, not in a knowledge base. They're the fastest to go stale and the most dangerous when they do.

2. **Add a schema/frontmatter to each entry** (e.g., `last-verified:` date, `category:` settled-decision | architecture-gotcha | ci-pattern) so CI can flag stale entries and agents can reason about validity.

3. **Tighten the README** — replace "treat with some skepticism" with clear validity rules: entries must describe stable facts (architecture constraints, CI patterns, settled decisions), not transient state. As rporres said, a wrong entry is worse than a missing entry.

4. **Add a CI check** — even a simple one that validates frontmatter schema and flags entries older than N months for re-review.

PR #610, opened 2026-07-18
Detach Ironic vmedia per-host as disk write completes during ABI

rporres:
Addresses the issue described in [gori-project/GoRI#937](https://github.com/gori-project/GoRI/issues/937): hosts whose
UEFI boot order ranks virtual media above disk re-enter the installer ISO
on the reboot that follows OpenShift being written to disk. Assisted Service
transitions them to installing-pending-user-action and blocks until a human
intervenes.

Root cause: the aggregate vmedia detach task ran after install-complete,
long after the per-host reboots that trigger the issue.

Rather than moving the bulk detach earlier (which would still leave a window
during disk write), this patch polls each host individually through the
Assisted Service infra-env per-host API and detaches Ironic vmedia the
moment that host's disk write is confirmed complete. Per-host watchers run
asynchronously in parallel with the install process.

Detach conditions derived from observed bare-metal/CI (virtual BMCs)
behaviour:

- Non-rendezvous host: AS status == "installed"
- Rendezvous/bootstrap host: status == "installing-in-progress" +
  status_info == "Rebooting" (disk write done, first reboot imminent)
- Fallback: 3 consecutive AS-unreachable polls — covers the rendezvous
  case where AS shuts down before it can report Rebooting

The Assisted Service credentials (cluster ID, URL, auth token) are needed
by both pre_install_validate_plugins.yaml and wait_for_deployment.yaml.
They are extracted into a shared task file (get_assisted_service_credentials.yaml)
included by both. wait_for_deployment.yaml guards the include with
`when: assisted_cluster_id is not defined` so credentials are only fetched
once in the normal flow; the guard also preserves standalone execution via
--tags wait-deployment.

## Summary
- **New Features**
  - Added a `wait-and-detach-vmedia` CLI subcommand that polls Assisted Service host status and detaches Ironic virtual media when appropriate.
  - Added dynamic Assisted Service credential discovery during pre-install validation.
- **Bug Fixes**
  - Virtual-media detachment now runs asynchronously per host, with playbooks waiting for completion and handling unreachable/skip conditions more robustly.
- **Chores / Tests**
  - Expanded test coverage and enhanced CI artifact collection to include vmedia fixture dumps.


playbooks/tasks/get_assisted_service_credentials.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+- name: Extract auth token from env file
+  ansible.builtin.set_fact:
+    _assisted_token_match: "{{ (r_rendezvous_env.content | b64decode) | regex_search('USER_AUTH_TOKEN=(.+)', '\\1') | default([]) }}"
+  no_log: true
+
+- name: Fail if auth token not found
+  ansible.builtin.fail:
+    msg: "USER_AUTH_TOKEN not found in /etc/assisted/rendezvous-host.env on rendezvous node"
+  when: _assisted_token_match | length == 0
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. Changed `| default([])` to `| default([], true)` so `regex_search` returning `None` (no match) is properly replaced by an empty list before the length check.

playbooks/tasks/get_assisted_service_credentials.yaml
```diff
@@ -0,0 +1,56 @@
...
+  until:
+    - r_assisted_clusters is success
+    - r_assisted_clusters.json | length > 0
+  no_log: true
+
+- name: Set Assisted Service credential facts
+  ansible.builtin.set_fact:
+    assisted_cluster_id: "{{ r_assisted_clusters.json[0].id }}"
+    assisted_service_url: "http://{{ rendezvousIP }}:8090/api/assisted-install/v2"
+    assisted_auth_token: "{{ _assisted_token_match | first }}"
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Not applicable. The Assisted Service runs on plain HTTP (port 8090) by design in Agent-Based Installer deployments — it is a local provisioning-network endpoint with no TLS support. Switching to HTTPS would require configuring TLS on the Assisted Service itself, which is outside the scope of this PR and not standard for ABI. The provisioning network is isolated and air-gapped, so the risk of token interception on that path is accepted.

playbooks/tasks/wait_for_deployment.yaml
```diff
@@ -1,4 +1,67 @@
...
+
+- name: Set ironic username fact
+  ansible.builtin.set_fact:
+    metal3_ironic_username: "{{ r_username_secret.stdout }}"
+  no_log: true
+
+- name: Set ironic password fact
+  ansible.builtin.set_fact:
+    metal3_ironic_password: "{{ r_password_secret.stdout }}"
+  no_log: true
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. Replaced `failed_when: false` on both secret reads with `failed_when: r_*_secret.rc != 0 or r_*_secret.stdout | length == 0`, so a missing or empty secret fails immediately with a clear error rather than silently proceeding with empty credentials.

src/enclave/tools/cli.py
```diff
@@ -209,9 +210,84 @@ def check_root_ca(config: str) -> None:
...
+def wait_and_detach_vmedia(
+    host_name: str,
+    assisted_service_url: str,
+    assisted_cluster_id: str,
+    assisted_auth_token: str,
+    ironic_base_url: str,
+    ironic_api_version: str,
+    ironic_user: str,
+    ironic_password: str,
+    poll_interval: int,
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. Changed `type=int` to `type=click.IntRange(min=1)` so zero and negative values are rejected at the CLI boundary before reaching `300 // poll_interval` or `time.sleep()`.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,171 @@
...
+    except requests.exceptions.ConnectionError:
+        return None
+    resp.raise_for_status()
+    hosts: list[dict[str, Any]] = resp.json()
+    for host in hosts:
+        if host.get("requested_hostname") == host_name:
+            return str(host.get("status", ""))
+    raise ValueError(
+        f"host {host_name!r} not found in Assisted Service cluster hosts response"
+    )
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. `get_host_status` now catches both `ConnectionError` and `Timeout` (previously `ReadTimeout` was not caught and would crash unhandled). The main loop tracks a `connection_error_streak` counter and only triggers detach after 3 consecutive unreachable polls — a single transient blip resets the streak on the next successful response.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,171 @@
...
+        auth=auth,
+        timeout=30,
+    )
+    if detach_resp.status_code not in {
+        requests.codes.no_content,
+        requests.codes.not_found,
+    }:
+        raise RuntimeError(
+            f"failed to detach vmedia for {host_name!r} (uuid={uuid}): HTTP {detach_resp.status_code}"
+        )
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. `detach_ironic_vmedia` now wraps: `ConnectionError`/`Timeout` on both the GET and DELETE requests as descriptive `RuntimeError`; `JSONDecodeError` from `node_resp.json()`; and `KeyError` for a missing `uuid` field — all with host and operation context included in the message.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,171 @@
...
+    Raises RuntimeError on timeout or unexpected Ironic failures.
+    """
+    logger.info(
+        "monitoring %s for disk write completion (interval=%ds, max=%d polls)",
+        host_name,
+        poll_interval,
+        max_polls,
+    )
+    info_every = max(1, 300 // poll_interval)
+    for poll_num in range(max_polls):
```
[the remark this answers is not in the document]
rporres: ✨ **Claude Code**: Fixed in 4943b3c. Replaced the poll-count loop with a wall-clock deadline: `start = time.monotonic()`, `deadline = start + max_seconds` (default 7200s), and a `while time.monotonic() < deadline` loop. Elapsed time is reported with `time.monotonic() - start`. The Ansible `async:` timeout is bumped to 7500s so the Python process has time to report its own timeout error rather than being SIGKILL'd.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,221 @@
...
+            host.get("hostname") == host_name
+            or host.get("requested_hostname") == host_name
+        ):
+            return str(host.get("status", ""))
+    seen = [f"{h.get('hostname')!r}/{h.get('requested_hostname')!r}" for h in hosts]
+    logger.debug(
+        "%s: not yet in host list — hosts seen (hostname/requested_hostname): %s",
+        host_name,
+        seen,
+    )
```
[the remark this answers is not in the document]
rporres: This is a debug log and INFO is the default logging severity. We need this in order to be able to properly understand what's going on in an error but it won't appear by default.

rporres:
@coderabbitai Is there anything missing to fix?

rporres:
✨ **Claude Code**: All three addressed in 3a4d7f4.

1. **`no_log: true` on async launch task** — restored. The original reason for removing it (it was hiding an `assisted_auth_token is undefined` error) is no longer applicable; that root cause was fixed by including `get_assisted_service_credentials.yaml` at the top of the task file. The `async_status` and debug tasks are unaffected — `no_log` suppresses Ansible's display of the task invocation, not the job file content that `async_status` reads.

2. **`async_status` retries budget** — changed to `retries: 375, delay: 20` (7500s total), matching the `async: 7500` worker timeout.

3. **`resp.raise_for_status()` and `resp.json()` uncaught** — both wrapped in `get_host_status()`: `HTTPError` becomes a `RuntimeError` with the HTTP status code, `JSONDecodeError` becomes a `RuntimeError` with context — both surface through the CLI's controlled error path.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _dump_hosts_response(host_name: str, hosts: list[dict[str, Any]]) -> None:
+    dump_dir = pathlib.Path(_FIXTURE_DUMP_DIR)
+    dump_dir.mkdir(parents=True, exist_ok=True)
+    fname = dump_dir / f"{host_name}_{int(time.monotonic() * 1000)}.json"
+    with fname.open("w", encoding="utf-8") as f:
+        json.dump(hosts, f, indent=2)
+
+
```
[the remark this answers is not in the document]
rporres: _dump_hosts_response is just to collect more fixtures, it will be removed before merging to main

src/tests/test_wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,361 @@
...
+    with pytest.raises(RuntimeError, match="HTTP 401"):
+        mod.get_host_status(SERVICE_URL, CLUSTER_ID, AUTH_TOKEN, HOST_NAME)
+
+
+def test_get_host_status_raises_on_invalid_json(requests_mock: Mocker) -> None:
+    """Raise RuntimeError when the Assisted Service response body is not valid JSON."""
+    requests_mock.get(HOSTS_URL, text="not-valid-json", status_code=200)
+    with pytest.raises(RuntimeError, match="not valid JSON"):
+        mod.get_host_status(SERVICE_URL, CLUSTER_ID, AUTH_TOKEN, HOST_NAME)
+
```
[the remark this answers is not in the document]
rporres: _dump_hosts_response is just to collect more fixtures, it will be removed before merging to main

rporres:
@coderabbitai please review and approve if you're ok or post your findings so that I can review/address them.

rporres:
✨ **Claude Code**: Addressed in def3d55

The streak fallback is intentionally absent from Phase 1. The two phases represent fundamentally different situations that cannot be treated symmetrically:

**Phase 1 — identification:** AS is unreachable because the rendezvous node is still booting from the ISO. The disk is blank at this point. Applying a streak-based detach here would remove the only boot path from a host that hasn't written anything yet — next reboot lands on an empty disk, bricking the installation. A loud timeout is the correct outcome when AS never comes up in Phase 1; if AS is permanently unreachable that early, the installation is broken regardless of vmedia state.

**Phase 2 — post-identification polling:** The streak models a specific, known event: AS shuts down *because* the bootstrap node finished writing to disk and rebooted, taking AS with it. By Phase 2 the host is already identified and actively installing, so "AS gone" carries semantic meaning — it is safe to infer that disk write completed and trigger detach.

Extending the streak to Phase 1 would trade a loud failure for a silent installation corruption. The asymmetry is intentional.

cc @coderabbitai

rporres:
@coderabbitai are we ok to approve it in the current state?

playbooks/tasks/get_assisted_service_credentials.yaml
```diff
@@ -0,0 +1,56 @@
...
+  no_log: true
+
+- name: Fail if auth token not found
+  ansible.builtin.fail:
+    msg: "USER_AUTH_TOKEN not found in /etc/assisted/rendezvous-host.env on rendezvous node"
+  when: _assisted_token_match | length == 0
+
+- name: Wait for Assisted Service cluster to be registered
+  ansible.builtin.uri:
+    url: "http://{{ rendezvousIP }}:8090/api/assisted-install/v2/clusters"
```
carbonin, line 38: Doesn't openshift-installer have some ABI specific commands to do things like this?

Maybe this one (`openshift-install agent wait-for bootstrap-complete`)? I'm not exactly sure what it does, but maybe it would be good enough or is that too late?
```
$ ./openshift-install agent wait-for --help
Wait for install-time events

Usage:
  openshift-install agent wait-for [flags]
  openshift-install agent wait-for [command]

Available Commands:
  bootstrap-complete Wait until the cluster bootstrap is complete
  install-complete   Wait until the cluster installation is complete

Flags:
  -h, --help   help for wait-for

Global Flags:
      --dir string         assets directory (default ".")
      --log-level string   log level (e.g. "debug | info | warn | error") (default "info")

Use "openshift-install agent wait-for [command] --help" for more information about a command.
```

playbooks/tasks/get_assisted_service_credentials.yaml
```diff
@@ -0,0 +1,56 @@
...
+  ansible.builtin.wait_for_connection:
+    timeout: "{{ wait_for_connection_timeout | default(120) }}"
+    sleep: "{{ wait_for_connection_sleep | default(5) }}"
+  delegate_to: rendezvous
+
+- name: Read Assisted Service env file from rendezvous node
+  ansible.builtin.slurp:
+    src: /etc/assisted/rendezvous-host.env
+  delegate_to: rendezvous
+  register: r_rendezvous_env
```
carbonin, line 20: This token should be available in the installer context directory. It's probably just as "suported" to get it from there as it is to pull it from the host so maybe that could save you an ssh?

playbooks/tasks/wait_for_deployment.yaml
```diff
@@ -1,7 +1,70 @@
...
+
+- name: Async wait for disk write complete and detach vmedia per host
+  ansible.builtin.command:
+    argv:
+      - enclave
```
carbonin, line 40: Just curious, but what are you using to determine when something should go in the CLI vs just in the playbooks?

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+    start = time.monotonic()
+    deadline = start + max_seconds
+    logger.info(
+        "monitoring %s for disk write completion (interval=%ds, deadline=%ds)",
+        host_name,
+        poll_interval,
+        max_seconds,
+    )
+
+    # Phase 1: identify this host in Assisted Service by requested_hostname.
```
carbonin, line 235: I'm not sure `requested_hostname` is always set.

Might be worth looking into the [logic assisted-service itself uses](https://github.com/openshift/assisted-service/blob/f46558c1df1388bc21448ae7fd3b249febb29798/internal/host/hostutil/host_utils.go#L47C1-L57C2) for hostname or using something hardware-based like MAC addresses if we're sure we have them.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+    - "unavailable": Assisted Service unreachable, caller handles streak
+
+    Detach conditions (from observed RHDP bare-metal behaviour):
+    - Non-rendezvous host: status == "installed"
+    - Rendezvous/bootstrap host: status == "installing-in-progress" and
+      status_info == "Rebooting" (disk write complete, bootstrap rebooting)
+    """
+    url = f"{assisted_service_url}/infra-envs/{infra_env_id}/hosts/{host_id}"
+    try:
+        resp = requests.get(url, headers={"Authorization": auth_token}, timeout=30)
```
carbonin, line 98: I think there's a python client if you'd rather use that.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+        status_info,
+    )
+    if status in {"error", "cancelled"}:
+        return "error"
+    if is_bootstrap:
+        if status == "installing-in-progress" and status_info == "Rebooting":
+            return "yes"
+    elif status == "installed":
+        return "yes"
+    return "no"
```
carbonin, line 130: Do we want to add a case here for if we missed the entire reboot and it's now `installing-pending-user-action`? We should be able to switch around the boot order at that point too.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+) -> None:
+    """Fetch the Ironic node UUID for host_name and DELETE its virtual media.
+
+    Accepts HTTP 204 (detached) and 404 (already detached or no vmedia).
+    Raises RuntimeError on any other response or transport failure.
+    """
+    headers = {"X-OpenStack-Ironic-API-Version": api_version}
+    auth = (ironic_user, ironic_password)
+    try:
+        node_resp = requests.get(
```
carbonin, line 148: I'd also expect there to be a pretty good ironic python client library

carbonin (commented):
This is one of the things we were trying to solve with the assisted+ironic "converged" flow for ACM-onprem installs. This way ironic itself is aware of when to reboot/change boot order.

It would be interesting to try to use the same pattern here, but I don't know if it's compatible with ABI.

src/enclave/tools/wait_and_detach_vmedia.py
dtantsur: Hey, somebody wrote a blog post on that! https://owlet.today/posts/easy-http-requests-with-openstacksdk/

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+    except (requests.exceptions.ConnectionError, requests.exceptions.Timeout) as exc:
+        raise RuntimeError(
+            f"cannot reach Ironic at {ironic_base_url} to detach vmedia for {host_name!r}: {exc}"
+        ) from exc
+    if detach_resp.status_code not in {
+        requests.codes.no_content,
+        requests.codes.not_found,
+    }:
+        raise RuntimeError(
```
dtantsur, line 188: If you don't reuse any existing clients, definitely retry HTTP 409 and 503.

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+            f"failed to detach vmedia for {host_name!r} (uuid={uuid}): HTTP {detach_resp.status_code}"
+        )
+    if detach_resp.status_code == requests.codes.not_found:
+        logger.info("vmedia for %s already absent (404)", host_name)
+    else:
+        logger.info("vmedia detached for %s", host_name)
```
dtantsur, line 194: The action is asynchronous and may fail without you noticing. Loop waiting for `GET f"{ironic_base_url}/v1/nodes/{uuid}/vmedia"`

src/enclave/tools/wait_and_detach_vmedia.py
```diff
@@ -0,0 +1,312 @@
...
+        return "error"
+    if is_bootstrap:
+        if status == "installing-in-progress" and status_info == "Rebooting":
+            return "yes"
+    elif status == "installed":
+        return "yes"
+    return "no"
+
+
+def detach_ironic_vmedia(
```
dtantsur, line 133: Ironic also has a boot device API https://docs.openstack.org/api-ref/baremetal/#set-boot-device we you should probably use after detaching to ensure you don't end up with something odd as your default boot device.

playbooks/tasks/get_assisted_service_credentials.yaml
rporres: We actually use that one in https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/wait_for_deployment.yaml#L8. We wait for that and then detach, but it looks it is too late for certain IBM hosts.

playbooks/tasks/get_assisted_service_credentials.yaml
rporres: I can't find it in an enclave working dir...

playbooks/tasks/wait_for_deployment.yaml
rporres: just complexity

src/enclave/tools/wait_and_detach_vmedia.py
rporres: ah, I see https://pypi.org/project/assisted-service-client/ . I'll take a look to see if it's worth the added dependency.

src/enclave/tools/wait_and_detach_vmedia.py
rporres: that makes total sense. In that case we would also need to reboot the host but it is a safeguard.

src/enclave/tools/wait_and_detach_vmedia.py
rporres: hmm, I was trying to avoid any boot device order modification... FWIW, we currently don't set anything after detaching vmedia: https://github.com/rh-ecosystem-edge/enclave/blob/main/playbooks/tasks/configure_hardware_ironic_detach_vmedia.yaml#L45

PR #614, opened 2026-07-22
Fix example certificates.yaml: comment out empty ironic HTTPS cert fields

maorfr:
## Summary
- `config/certificates.example.yaml` set `ironicHTTPSCertificate`/`ironicHTTPSKey` to empty strings (`""`) by default
- `schemas/certificates.yaml` enforces `minLength: 1` on both fields whenever they're present — so the example, if copied as-is, fails schema validation with `'' should be non-empty`
- This caused a real deployment failure reported in Slack (bare-metal deploy of `main-sovcloud-1784177656-46fdd18.tar.gz`), where the confirmed workaround was to omit the keys entirely rather than leave them empty
- Fixed by commenting the fields out by default (as placeholders), with a note that they must be omitted — not set to `""` — when Ironic HTTPS boot ISOs aren't in use

## Test plan
- [x] `make -f Makefile.ci validate-json-schema` passes
- [x] Manually confirmed `config/certificates.example.yaml` with these fields omitted validates cleanly against `schemas/certificates.yaml`

Related Slack threads:
- https://redhat-internal.slack.com/archives/C09G9BS83T9/p1784713443034469
- https://redhat-internal.slack.com/archives/C09MHJM91LG/p1784283912536639

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the example certificate configuration to leave optional HTTPS certificate and key settings unset by default.
  * Added guidance to configure these fields only when serving boot ISOs over HTTPS.
  * Clarified that a certificate must be provided when HTTPS boot-ISO serving is enabled.

### DIFF
--- a/config/certificates.example.yaml
+++ b/config/certificates.example.yaml
@@ -70,6 +70,15 @@ sslCACertificate: |
 # The SAN on this certificate must cover lzBmcHostname (DNS SAN, for Let's Encrypt certs)
 # or lzBmcIP (IP SAN, for private CA certs). Set lzBmcHostname in global.yaml when using
 # a publicly trusted certificate such as Let's Encrypt.
-ironicHTTPSCertificate: ""
-
-ironicHTTPSKey: ""
+#
+# Leave these commented out (not set to an empty string) when not serving boot ISOs
+# over HTTPS: the schema requires a non-empty value if the key is present at all.
+# ironicHTTPSCertificate: |
+#   -----BEGIN CERTIFICATE-----
+#   YOUR_IRONIC_HTTPS_CERTIFICATE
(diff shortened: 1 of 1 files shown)


agonzalezrh (approved):

rporres (approved):
yes, it is better this way.

maorfr:
✨ **Claude Code**: Addressed in 38eca33 — reworded the comment to accurately describe the schema contract (`ironicHTTPSCertificate` has an unconditional `minLength: 1`; `ironicHTTPSKey`'s non-empty requirement is conditional on the cert being set), instead of implying both fields are unconditionally required to be non-empty.

maorfr:
@coderabbitai approve

PR #643, opened 2026-08-05
update konflux pipelines from spec to ref

maorfr:
switching to use https://github.com/openshift/assisted-service/blob/master/.tekton/assisted-service-saas-main-pull-request.yaml

## Summary by CodeRabbit

* **Chores**
  * Updated pull request and push pipeline configurations to use the shared standard pipeline.
  * Standardized pipeline execution by referencing the main revision of the shared configuration.


maorfr:
let's try out the vanilla pipelines for a while

PR #422, opened 2026-06-02
Add --plugin option to operator-versions reconcile command

maorfr:
Add support for loading operator definitions from plugin descriptors via a new --plugin option. This allows reconciling operator versions for all operators defined in a plugin without manually specifying each operator's details.

The implementation extracts operator configuration loading into helper functions to reduce complexity and avoid code duplication between --use-defaults and --plugin code paths.

Assisted-by: Claude Code <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added `--plugin` option to the `operator-versions` command to load operator definitions from a plugin descriptor file.

* **Changes**
  * Option validation now enforces mutual exclusivity: `--plugin` cannot be used with `--name`, `--version`, `--namespace`, or `--csv-name`.
  * `--use-defaults` is now mutually exclusive with `--plugin` and direct flags.
  * Improved error messages for plugin and defaults YAML loading failures.

PR #688, opened 2026-08-11
Add --plugin option to operator-versions reconcile command

maorfr:
part of https://redhat.atlassian.net/browse/OSAC-3917

## Summary
- Adds `--plugin <name>` to `enclave reconcile operator-versions`, loading operator definitions from `plugins/<name>/plugin.yaml` instead of requiring `--name`/`--version`/`--namespace`/`--csv-name` to be passed manually.
  - Mutually exclusive with `--use-defaults` and the direct flags, same as the existing `--use-defaults` path.
  - Rejects plugins with `installOperators: false` and validates the plugin name against path traversal (`..`, `/`, `\`).
  - This reimplements #422 (closed, unmergeable — its branch predated the `src/` layout restructuring) on top of current `main`, carrying over the security hardening from that PR's CodeRabbit review: path-traversal validation on the plugin name and sanitized YAML parse error messages.
- Wires plugin operator reconciliation into the upgrade flow (`playbooks/upgrade.yaml`), since previously only `defaults/operators.yaml` operators were upgraded and plugin-installed operators (e.g. ODF, LVMS) were silently skipped.
  - New `playbooks/tasks/upgrade_plugins.yaml` (mirrors the discovery pattern in `tasks/deploy_plugins.yaml`) auto-discovers enabled plugins with `installOperators: true` and `installOperatorsFleet: false` — i.e. plugins whose operators are approved/upgraded directly on the management cluster rather than rolled out to fleet clusters via ACM policies — and runs `operator-versions --plugin <name>` for each.
  - Gated by the same `upgrade_operators` flag as the existing `--use-defaults` step.
- Updates `docs/UPGRADE.md` to document the new automated step and the manual per-plugin commands.

## Test plan
- [x] `make python-unit-test` — 245 passed, coverage 91.54%
- [x] `make python-linter-test` — ruff check + format clean
- [x] `make python-types-test` — mypy strict, no issues
- [x] New CLI tests: single/multi-operator plugin dispatch, mutual exclusivity with `--name`/`--use-defaults`, plugin not found, no operators defined, `installOperators: false`, path traversal (`..`, `/`, `\`)
- [x] `ansible-lint` (production profile) passes on `playbooks/upgrade.yaml` and `playbooks/tasks/upgrade_plugins.yaml`
- [x] `yamllint` passes on the same files

## Summary by CodeRabbit

* **New Features**
  * Added plugin-aware operator version reconciliation through the `operator-versions --plugin` command.
  * Upgrade workflows now automatically reconcile operator versions for eligible management-cluster plugins.
  * Added validation for plugin definitions, operator lists, installation settings, and unsafe plugin names.

* **Documentation**
  * Updated the upgrade guide with plugin operator upgrade behavior and manual commands for disabled automation scenarios.

* **Bug Fixes**
  * Added clearer handling for conflicting, missing, or invalid operator-version options.

### DIFF
--- a/src/enclave/reconcile/cli.py
+++ b/src/enclave/reconcile/cli.py
@@ -74,10 +74,16 @@ def _reconcile_operators_from_list(
     Each operator dict must have 'name', 'version', 'namespace', and
     optionally 'csvNames' (defaults to [name] when absent).
 
+    Validates all operator entries before reconciling any. If validation
+    fails for any entry, no reconciliation occurs.
+
     Raises:
         click.ClickException: If any operator entry is missing required
             fields or has invalid field types.
     """
+    # Validate all operators first, before reconciling any
+    validated_operators: list[tuple[str, str, list[str]]] = []
+
     for idx, op in enumerate(operators):
         # Validate operator entry structure
         if not isinstance(op, dict):
@@ -95,20 +101,23 @@ def _reconcile_operators_from_list(
         csv_names_raw = op.get("csvNames")
         if csv_names_raw is not None and (
             not isinstance(csv_names_raw, list)
-            or not all(isinstance(name, str) for name in csv_names_raw)
+            or not all(isinstance(name, str) and name.strip() for name in csv_names_raw)
         ):
             raise click.ClickException(
-                f"Operator entry {idx} has invalid 'csvNames': must be a list of strings"
+                f"Operator entry {idx} has invalid 'csvNames': must be a list of non-blank strings"
             )
 
         op_name = cast("str", op["name"])
         op_csv_names = cast("list[str] | None", csv_names_raw) or [op_name]
-        operator_versions_reconcile(
+        validated_operators.append((
             cast("str", op["version"]),
             cast("str", op["namespace"]),
             op_csv_names,
-            dry_run,
-        )
+        ))
+
+    # Reconcile only after all operators pass validation
+    for version, namespace, csv_names in validated_operators:
+        operator_versions_reconcile(version, namespace, csv_names, dry_run)
 
 
 def _load_defaults_operators() -> list[dict[str, object]]:
--- a/src/enclave/reconcile/cli.py
+++ b/src/enclave/reconcile/cli.py
@@ -26,6 +26,111 @@ def defaults_path(filename: str) -> Path:
     return path
 
 
+def plugin_descriptor_path(plugin_name: str) -> Path:
+    """Resolve the plugin.yaml path for a plugin name.
+
+    Plugins are not packaged into the built wheel, so this only resolves
+    against a repo checkout: src/enclave/reconcile/cli.py → src/enclave/ →
+    repo_root/plugins/<plugin_name>/plugin.yaml.
+
+    Raises:
+        click.ClickException: If plugin_name is empty or could escape the
+            plugins directory (contains '..' or a path separator).
+    """
+    if (
+        not plugin_name
+        or "/" in plugin_name
+        or "\\" in plugin_name
+        or ".." in plugin_name
+    ):
+        raise click.ClickException(
+            f"Invalid plugin name: {plugin_name!r}. Plugin name must be a "
+            "simple name without path separators or '..'."
+        )
+    enclave_pkg = Path(__file__).resolve().parent.parent
+    return enclave_pkg.parent.parent / "plugins" / plugin_name / "plugin.yaml"
+
+
+def _reconcile_operators_from_list(
+    operators: list[dict[str, object]], dry_run: bool
+) -> None:
+    """Call operator_versions_reconcile for each operator in a list.
+
+    Each operator dict must have 'name', 'version', 'namespace', and
+    optionally 'csvNames' (defaults to [name] when absent).
+    """
+    for op in operators:
+        op_name = cast("str", op["name"])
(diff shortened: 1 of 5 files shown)


maorfr:
✨ **Claude Code**: Addressed all CodeRabbit review feedback in 2e63284:

1. **Enhanced path validation** (`plugin_descriptor_path()`): Now resolves both the plugins root and descriptor path, then validates the descriptor remains relative to the plugins root using `.relative_to()` — prevents symlink escapes and Unicode normalization attacks.

2. **Operator entry validation** (`_reconcile_operators_from_list()`): Added comprehensive validation for each operator dict before accessing fields — requires non-empty string `name`/`version`/`namespace`, and if `csvNames` is present it must be a list of strings. Raises descriptive `ClickException` identifying the invalid entry.

3. **Multi-operator test improvement**: Changed `test_operator_versions_plugin_multiple_operators` to mock a plugin with **two** separate operator entries (not one with many CSVs), verifying `operator_versions_reconcile` is called exactly twice with the correct arguments for each. Added `test_operator_versions_plugin_real_odf` to preserve coverage of the single-operator-with-many-CSVs case.

4. **Additional test coverage**: Added tests for invalid operator entries (non-dict, missing name, invalid csvNames).

All checks pass: 249 tests, ruff, mypy, coverage 91.54%.

maorfr:
✨ **Claude Code**: Addressed all 5 CodeRabbit review comments in fb8b138:

## 1. 🚨 MAJOR - Validate before reconcile (lines 81-111, outside diff)
Refactored `_reconcile_operators_from_list()` to validate **ALL** operator entries before reconciling **any**. Previously, if validation failed on a later entry, earlier operators would already be reconciled (InstallPlans approved), leaving a partial upgrade. Now:
- Collects validated operators in a list during validation pass
- Reconciles only after the entire input passes validation
- Added `test_operator_versions_plugin_partial_validation_no_reconcile` to verify no reconciliation occurs when one valid + one invalid operator is passed

## 2. Reject blank CSV names (lines 94-105)
Enhanced csvNames validation to reject blank/whitespace-only strings: `name.strip()` check in the list comprehension. Prevents downstream reconciler from constructing invalid CSV resource names and timing out. Added `test_operator_versions_plugin_invalid_csvnames_blank` to cover this case.

## 3. Idempotent Ansible task (upgrade_plugins.yaml, lines 48-60)
Added `changed_when: "'[UPDATE]' in r_plugin_operator_versions_upgrade.stderr"` to the plugin operator upgrade task. The CLI logs `[UPDATE] namespace/op:version` when it approves an InstallPlan. Task now reports `changed` only when an actual upgrade occurred, not on every run.

## 4. Markdown linting (docs/UPGRADE.md, lines 49-62)
Added blank lines before/after fenced code blocks to satisfy MD031 markdownlint rule.

## 5. Ruff line limit (test_cli.py, lines 232-270)
Refactored long inline YAML fixtures to use adjacent string literals (parenthesized concatenation), keeping all lines under 88 chars.

---

**All checks pass**: 251 tests (+2 new), ruff, mypy, ansible-lint (production), coverage 91.57%.

src/enclave/reconcile/cli.py
```diff
@@ -26,6 +26,161 @@ def defaults_path(filename: str) -> Path:
     return path
 
 
+def plugin_descriptor_path(plugin_name: str) -> Path:
+    """Resolve the plugin.yaml path for a plugin name.
+
+    Plugins are not packaged into the built wheel, so this only resolves
+    against a repo checkout: src/enclave/reconcile/cli.py → src/enclave/ →
+    repo_root/plugins/<plugin_name>/plugin.yaml.
```
rporres, line 34: why not packaging them into the wheel as we do with operator defaults?

src/enclave/reconcile/cli.py
```diff
@@ -53,7 +208,11 @@ def cli(log_level: str) -> None:
     "--use-defaults",
     is_flag=True,
     default=False,
-    help="Load all operators from defaults/operators.yaml (mutually exclusive with --name, --version, --namespace, --csv-name)",
+    help="Load all operators from defaults/operators.yaml (mutually exclusive with --name, --version, --namespace, --csv-name, --plugin)",
+)
+@click.option(
+    "--plugin",
+    help="Load operators from plugins/<name>/plugin.yaml (mutually exclusive with --name, --version, --namespace, --csv-name, --use-defaults)",
```
rporres, line 215: I'd prefer a separate `reconcile` subcommand, since operator-versions looks very much related to `operators/`. Why not a `plugin-version` command to avoid the overload of the `operator-versions` one? This would also remove the mutual exclusivity logic. We could retain much of the current logic.

maorfr: thought about naming and structure a lot here. it's not strictly a plugin update, since operators are a part of a plugin, but update of other pieces is not yet determined.

rporres: I see. What about `plugin-operator-versions` subcommand then? What I don't like is that `operator-versions` is naturally bound with `operators/` directory and `defaults/operators.yaml` making things easy to understand without diving deep into it. You either use defaults or a single operator. Now we're introducing a complexity that makes it harder to understand (e.g. what defaults are used by `--use-defaults`, plugin or operators?) and maintain instead of the simpler:

* `operator-versions` handles things under `operators/`
* `plugin-operator-versions` handles things under `plugins/`

PR #726, opened 2026-08-24
OSAC-4379: update uv

rporres:


setup_ansible.sh
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$SCRIPT_DIR"
 
-UV_VERSION="0.11.13"
+UV_VERSION="0.12.5"
```
maorfr, line 8: perhaps this can be done using renovate?

maorfr (approved):

setup_ansible.sh
rporres: I don't know how, but I'll look into it 😄

rporres: https://github.com/rh-ecosystem-edge/enclave/pull/728

PR #770, opened 2026-09-04
OSAC-4908: Move helm and mirror-registry binaries to mirror.openshift.com

rporres:
## Summary

`developers.redhat.com` content-gateway is misbehaving/unreliable when downloading control binaries. This moves the `helm` and `mirror_registry` URLs in `defaults/control_binaries.yaml` to `mirror.openshift.com`, aligning them with `openshift_client` and `oc_mirror`, which already use it.

Checksums are unchanged — both mirror URLs return HTTP 200 and resolve (via the same `access.cdn.redhat.com` backend) to files matching the existing SHA256 values, so the binaries are identical.

## Ticket

OSAC-4908

## Summary by CodeRabbit

* **Chores**
  * Updated the download locations for the Helm and mirror registry binaries.
  * Binary checksums remain unchanged, preserving existing integrity verification.

### DIFF
--- a/defaults/control_binaries.yaml
+++ b/defaults/control_binaries.yaml
@@ -4,10 +4,10 @@ control_binaries:
     url: "https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.20.32/openshift-client-linux.tar.gz"
     checksum: "sha256:2deb2e225404047bdae2808036cf16e04612df57d2560a1152a01003f3f7a46d"
   helm:
-    url: "https://developers.redhat.com/content-gateway/file/pub/openshift-v4/clients/helm/3.17.1/helm-linux-amd64"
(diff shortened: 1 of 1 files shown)


javipolo (approved):

maorfr:
fixes https://github.com/gori-project/GoRI/issues/959

pgodowski:
can we ask the RH team to re-introduce the file in the existing location?
We have now customers who are trying to install and they cannot due to this binaries location change

maorfr:
> re-introduce the file in the existing location

what does this mean?

pgodowski:
> > re-introduce the file in the existing location
> 
> what does this mean?

I mean, to have helm cli available on developers.redhat.. site, i.e. the location we pointed initially

PR #790, opened 2026-09-08
Add OSAC global.services values

maorfr:
# Summary
- add `global.services` entries to the rendered OSAC chart values
- derive the enabled flags from `osacProfilesList`
- keep the existing values structure intact for backward compatibility

follow up on https://github.com/osac-project/osac/pull/684

## Summary by CodeRabbit

* **New Features**
  * Added global service configuration for CaaS, VMAAS, and BMAAS, enabled according to selected OSAC profiles.
* **Configuration**
  * Explicitly disabled the MAAS service.

### DIFF
--- a/plugins/osac/templates/values.yaml.j2
+++ b/plugins/osac/templates/values.yaml.j2
@@ -7,6 +7,17 @@
 cliImage: "{{ osac_images.cli }}"
 {% endif %}
 
+global:
+  services:
+    caas:
+      enabled: {{ ('caas' in osacProfilesList) | lower }}
+    vmaas:
+      enabled: {{ ('vmaas' in osacProfilesList) | lower }}
+    bmaas:
+      enabled: {{ ('bmaas' in osacProfilesList) | lower }}
+    maas:
+      enabled: false
+
 operator:
   image:
     pullPolicy: Always


plugins/osac/templates/values.yaml.j2
eurijon, line 43: Should we remove this once a new OSAC chart with https://github.com/osac-project/osac/pull/684 is released?

plugins/osac/templates/values.yaml.j2
```diff
@@ -7,6 +7,17 @@
...
+  services:
+    caas:
+      enabled: {{ ('caas' in osacProfilesList) | lower }}
+    vmaas:
+      enabled: {{ ('vmaas' in osacProfilesList) | lower }}
+    bmaas:
+      enabled: {{ ('bmaas' in osacProfilesList) | lower }}
+    maas:
+      enabled: false
+
```
eurijon, line 20: Is it possible to place the global section at the beginning of the file (before the `cliImage` configuration) to keep the same order as in the OSAC chart https://github.com/osac-project/osac/blob/main/osac-installer/charts/osac/values.yaml?

plugins/osac/templates/values.yaml.j2
eurijon, line 232: `bmf.enabled` has been removed from OSAC's `values.yaml` in https://github.com/osac-project/osac/pull/684/, and is now handled by the bmaas service, we'll need to adapt this section too.

eurijon: Will be addressed in a follow-up PR

eurijon: https://github.com/rh-ecosystem-edge/enclave/pull/797

plugins/osac/templates/values.yaml.j2
eurijon: https://github.com/rh-ecosystem-edge/enclave/pull/797

plugins/osac/templates/values.yaml.j2
eurijon: https://github.com/rh-ecosystem-edge/enclave/pull/797

PR #485, opened 2026-06-14
Prevent ACM from changing MCE subscription to Automatic

maorfr:
## Summary

- Adds `installer.open-cluster-management.io/mce-subscription-spec` annotation to MultiClusterHub
- Ensures MCE subscription remains with `installPlanApproval: Manual`
- For disconnected mode, also specifies the mirrored catalog source

## Problem

ACM's MultiClusterHub operator automatically changes the multicluster-engine subscription's `installPlanApproval` from `Manual` to `Automatic`, breaking the manual approval workflow required for disconnected environments. This was discovered during operator upgrade testing where MCE install plans were auto-approved, bypassing the `enclave reconcile` step.

See Slack thread: https://redhat-internal.slack.com/archives/C0B7CM99EH5/p1781260765092579

## Solution

Use the `installer.open-cluster-management.io/mce-subscription-spec` annotation ([docs](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.15/html/install/installing)) on the MultiClusterHub resource to instruct ACM to:
1. Keep `installPlanApproval: Manual` for MCE subscription
2. Use the mirrored catalog source (`cs-mirror-redhat-operators-v4-20`) in disconnected environments

## Test plan

- [ ] Verify MultiClusterHub is created with the annotation
- [ ] Verify MCE subscription remains `Manual` after ACM installation
- [ ] Test operator upgrade workflow with `enclave reconcile` for MCE
- [ ] Verify behavior in both connected and disconnected modes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Enhanced advanced cluster management to properly handle disconnected environment setups.
  * Improved subscription specification configuration for multi-cluster deployments.
  * Automatic installation plan approval management based on environment connectivity.
  * Refined cluster hub resource metadata annotations for advanced deployment scenarios.

PR #796, opened 2026-09-11
Set oadp-subscription-spec annotation on MCH for disconnected installs

nikhilgupta3:
Without this, the cluster-backup component's OADP subscription defaults to source: redhat-operators, which doesn't exist in air-gapped clusters and leaves MultiClusterHub stuck in Pending.

## Summary by CodeRabbit

- **New Features**
  - Added OADP subscription configuration alongside MCE when setting up Advanced Cluster Management.
  - OADP installation now supports disconnected environments and selects the appropriate subscription mode in connected environments.
  - MultiClusterHub setup now applies the OADP subscription configuration automatically.


maorfr:
this is similar to #485 but is missing explicit management of the operator as was done in #13. more information in https://github.com/gori-project/GoRI/issues/963#issuecomment-5659804812.

